PR #229, opened 2014-06-12
Remove "requiretty" in /etc/sudoers in fedora

oconnore:
"requiretty" prevents vagrant provisioners from becoming root during
provisioning, since the script runs without a tty.


cheeseplus:
Closing the PR as this has been resolved in the currently supported Fedora's (currently 21 & 22). Thanks!!

PR #321, opened 2015-02-12
Fix mac address issue for ens33 on centos 7

alappe:
Before I had an issue using the box with a static ip for NFS sharing in
vagrant, resulting in

```
ERROR: [/etc/sysconfig/network-scripts/ifdown-eth] Device ens33 has MAC
address 00:0C:29:F8:96:74, instead of configured address
00:0C:29:59:C4:97. Ignoring.
```

While the cleanup script already fixed this for `eth0`, CentOS7 seems to
have changed the device naming, therefor it's also needed for `ens33`.

Instead of removing the text of the entries, I opted for deleting the
whole line as there's no need for empty lines in the ifcfg-file.

### DIFF
--- a/packer/scripts/centos/cleanup.sh
+++ b/packer/scripts/centos/cleanup.sh
@@ -8,6 +8,10 @@ rm -f /tmp/chef*rpm
 # clean up redhat interface persistence
 rm -f /etc/udev/rules.d/70-persistent-net.rules
 if [ -r /etc/sysconfig/network-scripts/ifcfg-eth0 ]; then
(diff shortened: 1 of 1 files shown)


juliandunn:
The Ethernet devices are named in RHEL/CentOS >= 7 (and Fedora as well) based on the hardware and thus we can't just assume `ens33` is the one everyone will see. I like the idea, but it seems you should modify the patch to iterate through all files found in the `/etc/sysconfig/network-scripts` directory matching `ifcfg-*` and make the change that way.

alappe:
@juliandunn This should do it, I hope…

PR #327, opened 2015-02-24
Don't overwrite the vendor cacert.pem.

### LINKED ISSUE #318: Fedora kickstart script downloads CA cert bundle over HTTP
The kickstart scripts for Fedora (at least) download a new CA certificate bundle over unsecured HTTP from curl.haxx.se.

It seems to me that this should be done in some more secure fashion, if it is needed at all. Why does the CA cert bundle need to be updated, rather than using the one shipped by Fedora?

I have not reviewed the other build definitions to see if they also do this.

### LINKED ISSUE #325: need to get the correct cacert.pem for AWS on CentOS boxes
On CentOS, we download the cacert.pem bundle from the curl project.

The problem with this is explained in this mailing list post: http://curl.haxx.se/mail/archive-2014-10/0062.html

This manifests itself on CentOS when using Package Cloud for yum repositories, where, on the 5.11 box, it fails because SSL can't verify the certificate. Package Cloud has a valid SSL certificate, but their repositories are backed by AWS S3, so there's a redirect that happens, and SSL verification fails.

juliandunn:
Fixes #325 
Closes #318 

I tested the box build with CentOS 5 and 6 and it works fine

### DIFF
--- a/packer/http/centos-5.11/ks.cfg
+++ b/packer/http/centos-5.11/ks.cfg
@@ -69,8 +69,6 @@ yum
 -zd1211-firmware
 
 %post
-# update root certs
-wget -O/etc/pki/tls/certs/ca-bundle.crt http://curl.haxx.se/ca/cacert.pem
 # sudo
 echo "vagrant        ALL=(ALL)       NOPASSWD: ALL" >> /etc/sudoers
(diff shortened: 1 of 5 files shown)


juliandunn:
I'll take #328 over this one actually as it's more complete.

PR #328, opened 2015-02-24
Fixes #325, #318 - don't download cacert.pem

### LINKED ISSUE #325: need to get the correct cacert.pem for AWS on CentOS boxes

jtimberman:
This addresses both concerns of #318 and #325. We were downloading the
SSL CA bundle over http because at the point in time when we wanted to
even do that we might not have been in a state where the SSL
certificates from curl.haxx.se could be verified. Using http is just
as good at that point as using SSL without verification. However...

This addresses the concern raised in #325, whereby the upstream
cacert.pem removed certificates used by services such as AWS S3,
causing SSL connections to those sites to fail to verify. We should
rely on the ca-bundle.crt that comes with the openssl package on the
platforms in question (centos/fedora).

### DIFF
--- a/packer/http/centos-5.11/ks.cfg
+++ b/packer/http/centos-5.11/ks.cfg
@@ -69,8 +69,6 @@ yum
 -zd1211-firmware
 
 %post
-# update root certs
-wget -O/etc/pki/tls/certs/ca-bundle.crt http://curl.haxx.se/ca/cacert.pem
 # sudo
 echo "vagrant        ALL=(ALL)       NOPASSWD: ALL" >> /etc/sudoers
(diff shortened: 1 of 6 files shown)


jtimberman:
@juliandunn Here we go.

PR #336, opened 2015-04-02
Add centos 7.1

kramvan1:


bdwyertech:
Are y'all trying to maintain the eth0 interface naming convention, or are we going with the new method?  
Also, NetworkManager makes multiple interfaces with Vagrant a real PITA, since it increments the route metric, making all outbound traffic always flow through the default NAT'd interface.

irvingpop:
:thumbsup:  Tested this out,  works correctly.   Networking is the same as the centos-7.0 bento box, so no issues for me.

juliandunn:
@bdwyertech our general philosophy in Bento is to leave the defaults alone, so I guess that implies "new method".

juliandunn:
Merged in 757efdd8fc9a09a83a2730e135fb01247f27d4b2. Thanks!

PR #349, opened 2015-04-26
Add Ubuntu 15.04 vivid velvet

vincentaubert:
This PR adds Ubuntu 15.04 to bento

PR #350, opened 2015-04-26
Support Debian 8.0 (Jessie)

YpNo:
Support Debian 8.0 (Jessie), fix/force install MBR to the first device and add some user variables to allow packer user to set some own values


rickard-von-essen:
Works fine on **parallels** and **virtualbox** but fails on **vmware**. Might be same problem as in #349.

screen shot 2015-04-29 at 20 48 30

YpNo:
is not the 'ubuntu-server' option in "
tasksel tasksel/first multiselect standard, ubuntu-server" ?
I haven't vmware for now

YpNo:
I successfully build vmware box (with WMware PRO 7) and netinstall iso.
I retry with default values

YpNo:
Build too.
I have just an issue with gracefully halt detection. I was forced to restart and shutdown VM (and ACPI doesn't works because it freeze to '... Reboot: System Halted").
I looking about that

YpNo:
VMware box build much better with open-vm-tools. Standard VMware Tools try to use sysvinit to install his service. Need to wait VMware support Debian 8 and Systemd for classic install.
Otherwise, I have reproduced only once the issue during installation at this step. But I have found that crashed sometimes to different step. A new build and it goes

josephholsten:
@fnichol what would you need to get this merged? hugs? beverages? rebasing?

eriksencosta:
Maybe an advice like "the winter is coming!" :)

rmoriz:
"the winter is coming!"

fnichol:
Going to use #366 as the focus of this effort in an attempt to de-dupe some issues. Thanks!

PR #353, opened 2015-05-11
Mac OS X 10.10 support

juliandunn:
Used DMG prepared by Tim Sutton's scripts.

Ping @yzl @fnichol


fnichol:
Thanks! Will incorporate shortly.

PR #357, opened 2015-05-19
Add packer/bin/bento for building templates.

fnichol:
This is a first-step addition of a small binary wrapper around Packer to
help inject custom variables, inject metadata and re-normalize the
templates using `packer fix`.

In its present implementation, no external RubyGems are required in an
attempt to maximize portability between platforms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ packer_cache
 packer.log
 .DS_Store
 /packer-*/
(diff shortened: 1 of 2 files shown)


fnichol:
@schisamo, @yzl: After a rebase against master, this is also ready for review. First step towards build-time metadata.

PR #358, opened 2015-05-20
Remove VeeWee definitions from project

fnichol:
The remaining VeeWee definitions were exclusively Windows based but were
also suffering from age. The current state of Packer may allow for
WinRM-based templates for Windows in the near future but in the meantime
removing VeeWee simplifies this project's dependencies.


fnichol:
@schisamo, @yzl, Ready for a review. Kind of big, but I've been building these as a I go the last 2 days and confirming that the templates still build correctly.

schisamo:
:shipit: :sheep:

PR #361, opened 2015-05-26
Normalize Bento templates

fnichol:
This PR introduces a modified subcommand called `normalize` (formerly `fix`) which will validate each template and then runs it through `packer fix` to normalize the JSON output formatting. The subcommand creates a report at the end containing a list of all templates that were modified in the normalize execution, making this a candidate for running in CI.

### DIFF
--- a/bin/bento
+++ b/bin/bento
@@ -6,9 +6,10 @@ Signal.trap("INT") { exit 1 }
 $stdout.sync = true
 $stderr.sync = true
 
-require 'optparse'
-require 'ostruct'
-require 'benchmark'
+require "benchmark"
+require "digest"
+require "optparse"
+require "ostruct"
 
 class Options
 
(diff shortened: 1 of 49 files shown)


fnichol:
@yzl Ready for a quick review.

bin/bento
```diff
@@ -69,10 +70,10 @@ class Options
         },
         argv: templates_argv_proc
       },
-      fix: {
```
scotthain: I'm wondering if we should deprecate/warn but not remove the 'fix' CLI command in favor of making it an alias of 'normalize' at least for a little while. I may be being more conservative than is needed though. I feel like as small as it is, it is technically a 'breaking' change.

fnichol: At the moment, the `bin/bento` code has only been in master about a week (and still not documented). I could add an alias here, but was hoping it was under (or not) used.

scotthain: Ahhh gotcha - cool!

yzl:
lgtm

PR #362, opened 2015-05-26
Remove `chef_version` user variable & remove Chef installation option.

fnichol:
In an effort to simplify the potential for behavior proliferation that
user variables allows, the `chef_version` user variable is being
dropped. The default behavior was to *not* install Chef for all Bento
boxes as downstream tooling such as Test Kitchen or vagrant-omnibus can
handle Chef installations at boot time.


fnichol:
@yzl, @scotthain: Can either of you foresee the need the preserve the "install Chef" behavior in Bento?

scotthain:
:+1: I like the boxes agnostic :)

centos-5.11-i386.json
```diff
@@ -102,21 +102,18 @@
   ],
   "post-processors": [
     {
-      "output": "builds/{{.Provider}}/opscode_centos-5.11-i386_chef-{{user `chef_version`}}.box",
+      "output": "builds/{{.Provider}}/opscode_centos-5.11-i386.box",
```
lamont-granquist: These kinds of strings are hardcoded here:

https://github.com/test-kitchen/kitchen-vagrant/blob/master/lib/kitchen/driver/vagrant.rb#L112

Needs to still be "_chef-provisionerless.box"

lamont-granquist:
Yeah :+1:, keep these boxes simple and provisonerless.  We don't need to spend time supporting chef and puppet and ansible and salt and that entire bikeshed.

yzl:
:+1:

centos-5.11-i386.json
fnichol: Ah, that's a good catch. I'm hoping to update kitchen-vagrant just as soon as we can get boxes publishing up to Atlas. That could mean in the very short term that the box would have to be renamed when uploading to the S3 bucket. About 2 PRs down the road, I'm hoping to change the box file names even more so this is going to be important to keep in mind.

PR #363, opened 2015-05-27
WIP - Add initial support for Wind River Linux 5.1

ryancragun:
This is a WIP PR.  After upstream chef supports Nexus we can yank out the chef RPM bits and get this merged.  Ideally we could get upstream Vagrant to support Wind River and yank out the networking hacks.


cheeseplus:
Closing this as need seems to have passed and requires rebase.

PR #373, opened 2015-06-18
Minimized AutoYaST profiles for SLES 11.3/12 to enhance readability

mattiasgiese:
Some device specific parameters were removed becuse AutoYaST usually chooses the right default for drives/devices. This should make it easy to expand the list of potential builder (maybe add KVM/qemu as a target?)
I also went ahead and removed some unnecessary packages. Unfortunately i could only test the build with VirtualBox


cheeseplus:
Just wanted to bump this to let folks know this will be looked into - just needed to get some of the broader fixes and refactors in first.

cheeseplus:
Going to +1 and merge this and then we can deal with any fallout from the other builders as it arises as SLES is due for a re-factor anyway.

PR #386, opened 2015-07-02
updated debian's download redirector address

rmoriz:
see:
https://lists.debian.org/debian-devel-announce/2015/05/msg00003.html


fnichol:
Looks good, thanks!

PR #393, opened 2015-07-06
Adding Debian 8.1 links, removing Debian 8.0 templates

cheeseplus:
Debian major releases don't have point releases so 8.1 simply replaces 8.0. The 8.1 templates are already merged, this just updates the links (I've tested and pushed to S3) and removes the 8.0 definitions for clarity.


yzl:
:+1:

PR #396, opened 2015-07-09
[centos-*] Improve DRYness, correctness, and speed of CentOS templates.

fnichol:
The following improvements are made in this commit:

* Ensure that all related provisioner shell scripts are written using
  more conservative bourne code idioms
* Call all provisioner scripts with `sh -eux` which will (e) exit
  immediately if a command exits with a non-zero exit status, (u) treats
  unset variables as an error, and (x) outputs the command and its
  expanded arguments or associated word list
* Use the `template` user variable to replace repetition of `"centos-*"`
  style strings throughout
* Support a `HOME_DIR` environment variable for provisioner scripts to
  use
* Remove now-unused user variable `arch`
* Extract the `iso_checksum` and `iso_checksum_type` values into user
  variables and reference them in the templates
* Construct the `iso_url` value from the `mirror`, `mirror_directory`,
  and `iso_name` user variables
* Add a `ks_path` user variable and use it throughout the templates
* Add an `iso_name` user variable and use it throughout the templates
* Use the `http_proxy`, `https_proxy`, and `no_proxy` environment
  variables if set on the workstation and make them available to the
  provisioner shell scripts to speed up package installations, etc.


fnichol:
\cc @cheeseplus

cheeseplus:
:+1:

legal90:
Could you please explain why have you replaced all box names with this one?
```json
"box_basename": "__unset_box_basename__",
```

Now we always have to specify box name explicitly while running packer.

PR #399, opened 2015-07-10
Adding support for headless mode

cheeseplus:
Need headless mode to make builds from SSH sessions happy.


cheeseplus:
Parallels doesn't actually support headless mode via Vagrant but the option doesn't hurt anything currently so leaving it in case of future support.

fnichol:
We did our pair review, yay! :+1:

PR #400, opened 2015-07-11
Shortening vm_name to avoid Parallels box corruption

cheeseplus:
Due to this issue in the Go tar library https://github.com/mitchellh/packer/issues/1590 we are shortening VM names to just sneak under the bug limit. This only seems to affect Parallels but we're changing it across the board for the sake of consistency and de-duplication.


fnichol:
Paired and :+1:

PR #406, opened 2015-07-16
Rename centos images for clarity

patcon:
Perhaps I'm mistaken, but I CentOS minor versions are a little misleading, since the minor versions will automatically bump to the most recent available on `yum update`:
https://unix.stackexchange.com/questions/60660/minor-upgrades-and-version-numbers-on-centos

So even if someone downloaded as old 6.5 image, the normal update process brings them to 6.6. Thinking perhaps we could just say `centos-6` and `centos-7` for clarity.


legal90:
@cheeseplus, @fnichol What do you think about it?

patcon:
Last comment: Yes, it's arguably important, but it's akin to having an ubuntu slug of `ubuntu-14.04.2` (as I understand it :)

Simpler slug would mean you could just update the box in the atlas directory (with the same `centos-7` slug name), and use their box updater feature without being inconsistent

cheeseplus:
I'm generally onboard with this as I can't think of a downside and this maps to how the distro itself works. I'll chat with @fnichol and see if there is anything preventing us from making the switch.

yzl:
lgtm

cheeseplus:
So I reverted this merge for two reasons:

1. it was based off the 6.6 defintions
2. upon reflection we've found a great use case for not updating boxes

For 2 we realize that many people _want_ the old versions to test upgrades or have env restrictions. It's a difference between the reality of how people are constrained vs what version should be ideally used. We discussed making a slug for `centos-7` that points at latest but at the same time supporting MAJ.MIN as we already have and making it consistent across the distros.

PR #408, opened 2015-07-28
Add provisioning script for Xcode CLI tools

bgandon:
As I was building an OS X box to run Kitchen tests on a Chef cookbook,
I realized that even though Homebrew is the default package system for
Chef 12+, installing it actually requires the Xcode CLI tools to be already
there on the box.

Since the installation of those tools is somehow hackish (see how the
script `touch`es a file in `/tmp` in order to actually trigger an install through
the update system), it definitely makes sense to me that those tools are
provided on base boxes.

So I added Tim Sutton's script and it works pretty good.
Then I just wanted to share the result in case it might help others!

### DIFF
--- a/macosx-10.10.json
+++ b/macosx-10.10.json
@@ -178,6 +178,7 @@
       "scripts": [
         "scripts/common/metadata.sh",
         "scripts/macosx/hostname.sh",
+        "scripts/macosx/xcode-cli-tools.sh",
         "scripts/macosx/update.sh",
         "scripts/macosx/networking.sh",
         "scripts/macosx/vagrant.sh",
--- a/scripts/macosx/xcode-cli-tools.sh
+++ b/scripts/macosx/xcode-cli-tools.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+# Get and install Xcode CLI tools
+OSX_VERS=$(sw_vers -productVersion | awk -F "." '{print $2}')
+
+# on 10.9+, we can leverage SUS to get the latest CLI tools
+if [ "$OSX_VERS" -ge 9 ]; then
+    # create the placeholder file that's checked by CLI updates' .dist code
+    # in Apple's SUS catalog
+    touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
+    # find the CLI Tools update
+    PROD=$(softwareupdate -l | grep "\*.*Command Line" | head -n 1 | awk -F"*" '{print $2}' | sed -e 's/^ *//' | tr -d '\n')
+    # install it
+    softwareupdate -i "$PROD" -v
+    rm /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
+
+# on 10.7/10.8, we instead download from public download URLs, which can be found in
+# the dvtdownloadableindex:
+# https://devimages.apple.com.edgekey.net/downloads/xcode/simulators/index-3905972D-B609-49CE-8D06-51ADC78E07BC.dvtdownloadableindex
+else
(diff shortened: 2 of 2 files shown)


cheeseplus:
In building the bento boxes we are going for the least opinionated box possible and for adding build tools to OSX nodes there is opscode-cookbooks/build-essential.

We also don't add build tools for any other platforms so while this might be useful for a specific use case, it's not something general purpose and as such a bit outside of the scope of what Bento boxes are here to provide.

bgandon:
Indeed, Bento boxes are the least opinionated box possible, that's right. But you don't uninstall `apt` on Debian boxes either. Nor don't you uninstall `yum` on CentOS boxes just for the sake of being "least opinionated". The reason is that you need those to do anything automated on your box.

You might see the question as being about shipping any build tool, but it's not. It's about having a base package manager, because a Bento box is something meant for automation.

So, I don't see how the "least opinionated" argument would prevail here at the point it would disallow automation. Homebrew is not an opinion on OS X. It's a standard like `apt` and `yum` on their respective systems.

bgandon:
Maybe what's missing here in this PR is some more code that would go further and actually install Homebrew?

rickard-von-essen:
@bgandon IMHO there is a huge difference between ```yum``` on say CentOS and ```homebrew``` on OS X. I think the question to ask is: _Does this come with the most minimal install possible of this OS or is it a dependency of Vagrant, or is it needed by Packer to build the box?_ 

Yum -> yes
Homebrew/Xcode -> no

cheeseplus:
As @bgandon notes, Homebrew is not in the same class as apt or yum - homebrew, while the de facto package provider on OSX is not required for the building of the box nor is the official package manager of the OS. As Homebrew is also typically installed by a non-root user it is also often best to let the user take care of the install once they've provisioned their users.

Disallowing automation is not at all the case - we provide a reasonably minimal base upon which automation can take place - i.e. using the `homebrew` cookbook which pulls in `build-essentials`. It's not that I am entirely opposed to the suggestion but we are indeed wading out into "opinion" territory.

jtimberman:
Hello!

Homebrew is the default provider for the `package` resource on OS X in Chef 12 because it's more commonly installed on OS X systems than macports. Prior to Chef 12, macports was the default provider for `package`. The package provider used by Chef is orthogonal to whether Xcode should be installed on the base boxes built with bento templates, however.

To be clear: We should *not* install Xcode. Specific reasons why not:

Most visible to users is that the installation of Xcode is huge. Adding that to the box build time can be an annoyance. Note that by Apple's license terms, we cannot make OS X base boxes publicly available. This means that users will need to build their own base boxes anyway.

Another issue is that Xcode is proprietary software that requires license agreement by the user for installation. It is assumed that when someone purposefully uses the `build-essential` cookbook, they're agreeing to the license for the software it installs. Yes, Xcode has open source components, and those are actually the ones that tools such as macports and homebrew actually care about. However, we can't assume the motivation of users to have Xcode is solely for macports or homebrew. Maybe they want to build iOS CI nodes? Who knows. 

That brings up another point. The reason to have base boxes of OS X are many and varied. Some users won't care about having Xcode, because they're not doing anything with these boxes that involves homebrew. We can't know, and as I said, the user will need to build their own base boxes. It's much easier to build a box without Xcode by default and add it in later, e.g., start minimal box, run build-essential against it, make a new box out of that. 

Finally, Xcode is not part of the operating system; it's not a dependency for OS X to function, or to install other software. Nor are macports or homebrew. This is in stark contrast with `apt` and `yum` which are core to their respective  platforms, because they're how all other software is installed on the systems. Yes, OS X has a built in package manager, but it's generally regarded as difficult to use, and it doesn't interact over a network, or resolve dependencies for installation. In fact, it's pretty much the reason why tools like macports and homebrew exist. :smile:.

bgandon:
You're all stressing very good points indeed. Thanks for the detailed answers then.

PR #411, opened 2015-07-29
Add RHEL 6.5 configuration.

markovuksanovic:


tas50:
This probably won't result in a 6.5 host since the Yum update that is run during build is going to pull in all the latest packages.

cheeseplus:
Correct @tas50 with CentOS/RHEL there is only the latest minor release because of how yum update works. Closing.

PR #413, opened 2015-08-07
[macosx-*] Increase disk_size to ~40G from ~20G.

fnichol:
When running some tests of the Xcode and Vagrant cookbooks, I found that the virtual machines were running out of drive space before the Chef runs completed. This should buy the VMs a bit more headroom.


yzl:
:+1:

legal90:
:+1:

PR #414, opened 2015-08-07
Opscode -> Chefs and other minor stuff

tas50:


cheeseplus:
This is easy and unobjectionable +1

PR #367, opened 2015-06-03
fedora-22 support

joemiller:
fedora-22 boxes.

- fedora-22-x86_64:  built OK on virtualbox. Don't have vmware or parallels to test.
- fedora-22-i386: this does not build. Instead I get the following error. Omitting the `--pesize=32768` from the ks.cfg doesn't help either, it just changes the message from 32 MiB to 4 MiB. Not sure what's up with this one and can't find anything from some brief google searching, and we only need the 64b box.

image

PR #418, opened 2015-08-12
Add Fedora 22 boxes

rickard-von-essen:
Note: added package ```dnf-yum``` to kickstart file to be explicit.

### DIFF
--- a/fedora-22-i386.json
+++ b/fedora-22-i386.json
@@ -1,0 +1,143 @@
+
+{
+  "builders": [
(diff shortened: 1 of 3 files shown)


cheeseplus:
As reported in https://github.com/chef/bento/pull/367 it looks like 386 is potentially busted but I'm happy to merge and just not publish the 386 box if/when it can be resolved.

cheeseplus:
+1

PR #424, opened 2015-08-19
Freebsd 10.2

geoffgarside:
Update FreeBSD 10.1 manifests to 10.2.

Change the freebsd-update part as it now supports a `--not-running-from-cron` argument to be used with automated tools. We check the major_version to see if it *should* be supported or not.

I've also added a note in that the current update.sh will fail when the FreeBSD release being installed doesn't have any updates available as the `freebsd-update install` will return an errcode 1 which will cause the shell (running with `-e`) to bail out.

If you'd rather the commits were squashed let me know and I can sort that.


cheeseplus:
This is great @geoffgarside as I was encountering this while testing a build pipeline. I think the right approach here is to make an additional commit to catch and hanlde an already up-to-date install (no updates available)

geoffgarside:
I'm not quite sure how to test this now. FreeBSD 10.2 now has a p1 set of updates. What I tried however was this

```sh
env PAGER=/bin/cat $freebsd_update fetch | tee /tmp/freebsd-update-fetch.log

tail -n1 /tmp/freebsd-update-fetch.log | grep "No updates needed"
if [ $? -eq 0 ]; then
  env PAGER=/bin/cat $freebsd_update install
fi
```

unfortunately if there are updates available this screws up as it appears to execute the tail/grep in the middle of `freebsd-update` emitting the list of changes which are going to be applied. In writing it out again I'm wondering if perhaps the PAGER might be causing the problems as the fetch might have closed its stdout when the PAGER gets its content to print out.

I'm happy to try and fix it as part of this PR, but I'm a bit lost on how to actually fix it...

fnichol:
Huh, didn't think we'd get a less-automatable system update here :crying_cat_face:

geoffgarside:
I think the previous versions of `freebsd-update` also return an error when running `install` without any updates as well. So that part I don't think has changed. We've just got the `--not-running-from-cron` option now to let the `fetch` run without complaining that its lacking a TTY.

Its annoying that theres no good way to tell from the return code of `fetch` if theres any updates that are needed. It might make sense for `install` with `--not-running-from-cron` to not complain about no updates, but it technically complains that you've not run fetch as it can't see any updates to apply.

cheeseplus:
In the interest of functioning > perfect I'm giving this a +1 as it allows us to build 10.2

PR #442, opened 2015-09-11
Setting basename to a more sensible default

### LINKED ISSUE #433: Building boxes results in files with unhelpful names
I'm using the latest version of Bento from git and finding that if I build a box like so, following the instructions:

```
packer build -only=virtualbox-iso -var 'mirror=...' centos-7.1-x86_64.json
```

the result is that the box is named ``builds/__unset_box_basename__.virtualbox.box``.  This isn't very helpful, especially when I was building two boxes from different OSes at the same time.

I note that the string ``__unset_box_basename__`` is hard-coded into most, but not all, provided JSON files in Bento and that the comment at https://github.com/chef/bento/issues/430#issuecomment-135843675 mentions how to configure the box basename.

If that comment is the prescribed the way to avoid the above problem, then may I suggest documenting this process in the README?  Otherwise, can a more useful name be set in each JSON file?

cheeseplus:
Fix #433

### DIFF
--- a/centos-5.11-i386.json
+++ b/centos-5.11-i386.json
@@ -136,7 +136,7 @@
     }
   ],
   "variables": {
-    "box_basename": "__unset_box_basename__",
+    "box_basename": "centos-5.11-i386",
     "build_timestamp": "{{isotime \"20060102150405\"}}",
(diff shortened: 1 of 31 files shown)


yzl:
:+1:

PR #447, opened 2015-09-14
Updating platforms to use sha256

cheeseplus:
Other platforms using sha1 will update this in their upcoming respective refactor PRs


yzl:
lgtm

PR #452, opened 2015-09-17
Remove the headless option for Parallels builder.

rickard-von-essen:
This option will not be allowed anymore by packer,
it has never been documented and never hade any effect.

See https://github.com/mitchellh/packer/pull/2662


cheeseplus:
Looks good

PR #456, opened 2015-09-23
remove duplicate auth/authconfig and use sha512 algorithm for password hashing

mmckinst:
auth and authconfig are the same thing, they're used twice in the kickstart
files. the first one tells authconfig to use sha512, the second one tells it to
use md5. the end result the machine is configured to use md5 instead of sha512
for password hashing

https://github.com/rhinstaller/pykickstart/blob/master/docs/kickstart-docs.rst#auth-or-authconfig


cheeseplus:
Seems completely sensible to me +1

PR #457, opened 2015-09-24
bento command needs two dashes

mmckinst:


cheeseplus:
Nice catch - was fixing that in another PR but happy to accept this one.

PR #374, opened 2015-06-19
squashed commit to add all the qemu options on all the systems.

dmlb2000:
This adds the qemu options for all systems. Should be "squashed" I didn't use the git features for this I just did a diff and committed it, it was easier to inspect the diff that way.

PR #461, opened 2015-09-27
squashed commit to add qemu options for all boxes

dmlb2000:
This is a new squashed commit for adding the qemu options for all commits, this was pull request #374

And hopefully passing travis tests.


dmlb2000:
Here's everything I've been able to build at home with these.

centos-5.11
centos-5.11-i386
centos-6.7
centos-6.7-i386
centos-7.1
debian-6.0.10
debian-6.0.10-i386
debian-7.9
debian-7.9-i386
fedora-21
fedora-21-i386
fedora-22
freebsd-10.2
freebsd-9.3
freebsd-9.3 (i386)
opensuse-13.2
opensuse-13.2-i386
ubuntu-10.04
ubuntu-10.04-i386
ubuntu-12.04
ubuntu-12.04-i386
ubuntu-14.04
ubuntu-14.04-i386
ubuntu-14.10
ubuntu-14.10-i386

I'm not sure what's more or less supported than others. The ones that are missing from this list that I have in the pull request are

debian-7.8* (cd missing from default location)
debian-8* (either grub problem or cd not ejecting, reboots right back into the install CD)
fedora-22-i386 (odd partitioning issue that works on x86_64 but not on i386, seems like this is a bug in anaconda)
freebsd-10.2-i386 (the VM locked up hard when doing the clean.sh script)
oracle* (no local install CD)
rhel* (no local install CD)
sles* (no local install CD)
ubuntu-15.04* (similar issue to debian-8*)

cheeseplus:
Thanks for the rebase! I can confirm the issues with debian-7.8, fedora-22-i386, freebsd-10.2-i386 (perhaps 9.3-i386) exist across the board so not much worries there. Given that we aren't going to be able to build/ship/support these ourselves I'm comfortable merging these given the smaller audience.

cheeseplus:
Before I merge, is there a need to add lines like the following:
https://github.com/dmlb2000/bento/commit/169f547c38d9439c0c6e4dd0961d7a2fe3c3fd9a#diff-0bd6af31ad96430ffd60f5c0296fb980R165 for the other builders?

dmlb2000:
I tried not to break the other builds with the first commit. Then figured out that using disk_interface as scsi didn't seem to let grub install properly on debian 8* and ubuntu 15.04*. I then removed it and verified that using virtio as the backend device for those OSs seems to work better, grub installs and the box gets built properly. However, debian 8* and ubuntu 15.04* are using some version of debian-installer that requires to preseed the grub install device instead of just using a sane default. So that makes that preseed variable dependent on the type of virtualization you are using, vda for virtio and sda for most others.

cheeseplus:
So after another review I'm going to merge and if anything breaks I'll fix it but this all seems pretty great :+1:

PR #465, opened 2015-10-02
Fixing Fedora 22 build tools to fix vm tools

cheeseplus:
Fedora 22 builds are broken because apparently one cannot install GCC from kickstart (reasons unknown)

- add build tool script just for Fedora 22
- update kickstart and remove deprecated `--nobase` option
- doesn't fix i386 build but that is it's own yak


yzl:
:+1:

PR #466, opened 2015-10-03
Adding OmniOS r151014

cheeseplus:
More Illumos for the Illumos Gods

### DIFF
--- a/omnios-r151014.json
+++ b/omnios-r151014.json
@@ -1,0 +1,235 @@
+{
+  "builders": [
+    {
+      "boot_command": [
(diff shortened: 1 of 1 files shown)


yzl:
:+1:

fd1e4020:
Has there been an upstream refresh of the ISO image? The published checksum doesn't match the json's.

rickard-von-essen:
This seems to happen from time to time, see #470

cheeseplus:
Yea, the OmniOS folks keep the same image name but change the ISO which is incredibly annoying and a less than ideal release engineering practice :/

PR #475, opened 2015-10-13
[DRY] up rhel

cheeseplus:
- Add RHEL 6.7
- DRY up all RHEL


cheeseplus:
Won't build currently because this line https://github.com/chef/bento/blob/master/scripts/centos/cleanup.sh#L5 fails due to lack of RHN subscription.

cheeseplus:
Skipping `yum clean all` on RHEL as you can't run any yum commands without having an RHN subscription.

PR #479, opened 2015-10-19
Support swapless builders in minimize.sh

### LINKED ISSUE #420: minimize.sh fails with no swap partition
With set -e this line will exit minimize.sh if we got no swap partition, seems like that wasn't the intention ...

https://github.com/chef/bento/blob/master/scripts/common/minimize.sh#L3

sheldonh:
The blkid utility returns exit status 2 when the requested label can't
be found.

Fixes #420


cheeseplus:
Looks good to me!

PR #483, opened 2015-10-28
Ubuntu 15.10 support

rickard-von-essen:
Tested on Parallels/VirtualBox/VMware, but not Qemu.

- VMware fails to install the VMware tools with VMware Fusion Pro 7.1.2.
-  Ubuntu text mode installer is broken and the ```boot_command``` from the workaround in https://github.com/boxcutter/ubuntu/issues/22#issuecomment-148531379 is needed. See [Ubuntu bugs - debian-installer #1505839](https://bugs.launchpad.net/ubuntu/+source/debian-installer/+bug/1505839) and boxcutter/ubuntu#22.


rickard-von-essen:
[Ubuntu bugs - debian-installer #1505839](https://bugs.launchpad.net/ubuntu/+source/debian-installer/+bug/1505839) is probably not going to be fixed and released anytime soon so I think it's best sticking with the workaround. 

Regarding VMware Fusion support it would be nice if someone could test it on Fusion 8, if it works there I would say LGTM.

cheeseplus:
Another solid one @rickard-von-essen <3

PR #484, opened 2015-10-31
Add template for OS X 10.11 El Capitan

legal90:
Project [osx-vm-templates](https://github.com/timsutton/osx-vm-templates) already supports OS X 10.11 El Capitan.

Here I've just copied the template `macosx-10.10.json` and changed the name inside.
It works fine - verified with `parallels` provider for Vagrant.

### DIFF
--- a/macosx-10.11.json
+++ b/macosx-10.11.json
@@ -1,0 +1,230 @@
+{
+  "builders": [
+    {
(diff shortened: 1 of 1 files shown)


cheeseplus:
Will give this a look this week - I seem to remember some other providers having issues.

PR #485, opened 2015-10-31
Retire Ubuntu 10.04, which is end-of-life.

juliandunn:
https://lists.ubuntu.com/archives/ubuntu-announce/2015-April/000196.html


cheeseplus:
So my thinking is that we'll create an archive folder of some sort for old definitions because even if it's EOL, folks are still going to want to build it anyway Because Reasons. We won't maintain builds in CI but it at least makes it easier then checking out an old version of the repo.

cheeseplus:
Closing this PR for the reasons listed above but will move to an archive folder as stated.

PR #487, opened 2015-11-04
attempt to manage sshd_config in an [more] idempotent manner

jhoblitt:
Blindly appending entries to `sshd_config` is problematic if the
configuration keys are already defined and/or the configuration file is
missing a trailing newline.


cheeseplus:
:+1:

jhoblitt:
:smiley:

Issue #488, opened 2015-11-04
Fedora 22 & 23 i386 box not possible to build


rickard-von-essen:
Kickstart of Fedora 22 & 23 i386 boxes fails Kickstart due to https://bugzilla.redhat.com/show_bug.cgi?id=1264571 
I'll close this when the upstream bug is resolved.

rickard-von-essen:
PR #418 is affected by this issue.

cheeseplus:
Yep I think that even 21 has an issue too for i386 when I try to build as well.

cheeseplus:
This shows closed so closing here too as Fedora 22 is EOL - we can open a separate issue if 23 is still affected. https://bugzilla.redhat.com/show_bug.cgi?id=1264571

PR #489, opened 2015-11-04
Support Fedora 23

rickard-von-essen:


rickard-von-essen:
i386 boxes won't build due to #488

rickard-von-essen:
Also the VMware box fails to install vmware-tools under VMware Fusion 7.

rickard-von-essen:
Someone with access to Fusion 8 who could test if the VMware tools install succeeds there.

jhoblitt:
#488 didn't block fedora 22 i386 being merged ( #418 ) so I don't think it should hold up f23 either.  Alternatively, i386 could be split into a separate PR and x86_64 could be merged.

PR #492, opened 2015-11-11
FreeBSD 11.0 support

rickard-von-essen:
Changes compared with 10.2:
1) ~~http/freebsd-11.0/install.sh _start_ and _size_ have been adjusted to work:~~

``` diff
< gpart add -b 34 -s 94 -t freebsd-boot $DISKSLICE

---
> gpart add -b 40 -s 88 -t freebsd-boot $DISKSLICE
```

2) Switched to bsdinstall

TODO:
- [x] `minimize.sh` doesn't finish.


scripts/freebsd/update.sh
```diff
@@ -19,8 +19,10 @@ fi
 
 # Update FreeBSD
 # NOTE: this will fail if there aren't any patches available for the release yet
-env PAGER=/bin/cat $freebsd_update fetch;
-env PAGER=/bin/cat $freebsd_update install;
+if [ "$major_version" -lt 11 ]; then
+  env PAGER=/bin/cat $freebsd_update fetch;
+  env PAGER=/bin/cat $freebsd_update install;
+fi
```
tas50: is freebsd_update deprecated or is this gated just to work around the failures when there's no new updates?

rickard-von-essen: No this was just a quick work around that update fails when there are no patches to install. I have done a proper solution for [boxcutter/bsd](https://github.com/boxcutter/bsd/blob/master/script/freebsd/update.sh#L22). I'll port that. 

Thanks for reminding me.

cheeseplus:
Can we get this one rebased by chance? Release is right around the corner :D

rickard-von-essen:
I was just about to rebase to the 11.0 beta 1. But I'll think I'll rework it a bit to include my proposal from #558. Any comments on that is very welcome.

rickard-von-essen:
This should be ok now, just waiting for the _RELEASE_.

cheeseplus:
According to the schedule https://www.freebsd.org/releases/11.0R/schedule.html

RELEASE builds began on 28 September 2016
RELEASE Announcement should be October 5, 2016

Do we want to go ahead and start pushing?

rickard-von-essen:
The release is not out yet, it was delayed.  There was a email some days
ago saying the final build was on the way.

rickard-von-essen:
See
https://lists.freebsd.org/pipermail/freebsd-announce/2016-September/001752.html

rosstimson:
Looking forward to seeing this merged now that FreeBSD 11 is now officially released.

rickard-von-essen:
Updated to _FreeBSD 11.0 RELEASE_

scripts/freebsd/update.sh
```diff
@@ -19,7 +19,7 @@ fi
 
 # Update FreeBSD
 # NOTE: the install action fails if there are no updates so || true it
-env PAGER=/bin/cat $freebsd_update fetch;
+env PAGER=/bin/cat $freebsd_update fetch || true;
```
rickard-von-essen: 11.0 fails on fetch. I guess it's because there is no mirrors yet. Didn't have time to dig into it.

drq883:
You don't even need the -b option on the gpart add for freebsd-boot. I just removed it and created a vagrant image just fine.

Also I used these options on the dd command to make it finish properly:
  bs=1m count=61440
which only copies 60G of data. The disk is 40G in my json file.

scripts/freebsd/vmtools.sh
```diff
@@ -44,7 +44,7 @@ vmware-iso|vmware-vmx)
     mdconfig -a -t vnode -f $HOME_DIR/freebsd.iso -u 0;
     mount -t cd9660 /dev/md0 /tmp/vmfusion;
     tar xzf /tmp/vmfusion/vmware-freebsd-tools.tar.gz -C /tmp/vmfusion-archive;
-    /tmp/vmfusion-archive/vmware-tools-distrib/vmware-install.pl --force-install;
+    /tmp/vmfusion-archive/vmware-tools-distrib/vmware-install.pl --default;
```
cheeseplus: Probably need to add a check here like we have in other scripts for the vmware version and set the option as appropriate.

cheeseplus: Here are some easy lines to copy/pasta - https://github.com/chef/bento/blob/master/scripts/common/vmware.sh#L18-L23

rickard-von-essen:
Fixed support for both VMware Fusion 7 and 8 (+).

cheeseplus:
I know this started before we implemented DCO (as you can see from the bot) but if you could repush with signed commits (--signoff via git) then I can get this merged. Or I can do it myself and credit you. Either way, let me know and we'll get this merged!

rickard-von-essen:
Fixed the DCO.

cheeseplus:
lgtm

PR #494, opened 2015-11-21
fix or suppress all shellcheck warnings

jhoblitt:
This change-set represent the minimum possible change to either fix or
suppress all warnings from the shellcheck linter.  The presence of these
warnings is an annoyance for those with automatic linting enabled in
their editor of choice.

This include dead variable removal but no logic or stylistic
improvements.

### DIFF
--- a/scripts/centos/cleanup.sh
+++ b/scripts/centos/cleanup.sh
@@ -1,12 +1,12 @@
 #!/bin/sh -eux
 
 if [ -s /etc/oracle-release ]; then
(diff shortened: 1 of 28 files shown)


cheeseplus:
I've got no objections here, love it +1

jhoblitt:
@cheeseplus Great!  What do you think about adding shellcheck to travis?  The downside to this is that the package has vanished from the travis apt mirrors and sudo has to be enabled to use it.

PR #495, opened 2015-11-22
Fix grep character class syntax in sshd.sh script

legal90:
This PR fixes the following error:
```
==> parallels-iso: Provisioning with shell script: scripts/common/sshd.sh
    parallels-iso: grep: character class syntax is [[:space:]], not [:space:]
    parallels-iso: grep: character class syntax is [[:space:]], not [:space:]
```

cc: @jhoblitt


cheeseplus:
Looks good

jhoblitt:
Nice catch! My bad :(

PR #497, opened 2015-11-24
Disable automatic udev rules for network interfaces in CentOS

legal90:
This PR makes thec lean-up of udev rules similar to ubuntu/debian approach:
https://github.com/chef/bento/blob/6d5771a/scripts/ubuntu/networking.sh#L3-L8
https://github.com/chef/bento/blob/250efaa/scripts/debian/networking.sh#L3-L8

It's not enough just to remove `/etc/udev/rules.d/70-persistent-net.rules`. We should also replace it with a directory and remove `/lib/udev/rules.d/75-persistent-net-generator.rules` to prevent the generating of new rules. 

P.s. It is already implemented in boxcutter project and it works fine:
https://github.com/boxcutter/centos/blob/a732c24/script/cleanup.sh#L3-L14

### DIFF
--- a/scripts/centos/cleanup.sh
+++ b/scripts/centos/cleanup.sh
@@ -19,6 +19,9 @@ fi
 
 # Clean up network interface persistence
 rm -f /etc/udev/rules.d/70-persistent-net.rules;
+mkdir -p /etc/udev/rules.d/70-persistent-net.rules;
+rm -f /lib/udev/rules.d/75-persistent-net-generator.rules;
+rm -rf /dev/.udev/;
 
 for ndev in `ls -1 /etc/sysconfig/network-scripts/ifcfg-*`; do
     if [ "`basename $ndev`" != "ifcfg-lo" ]; then


legal90:
It really makes sense for `parallels` provider. Currently after `vagrant up` the guest OS generates a new udev rules for network devices. Everything works fine, but if user decide to pack this env to the custom box (with `vagrant package`), then such box will be packed with these pre-created rules.
This box is unusable: network devise order is shifted (now it starts from eth1 instead of eth0) and `vagrant up` fails with timeout:
https://github.com/Parallels/vagrant-parallels/issues/218

It's really inconvenient to remove `/etc/udev/rules.d/70-persistent-net.rules` manually before packaging. It is better just to prevent new rules generation here, while building the base box with Packer.

oskarpearson:
:+1: 

This bit both me and a colleague when trying to build packer-based configs using this repo as a base.

Oskar

cheeseplus:
+1

PR #498, opened 2015-11-24
Fix dhcp settings cleanup in Ubuntu 12.04 and higher

legal90:
There is a typo. It should be `/var/lib/dhcp/*`, like in the similar script for Debian: https://github.com/chef/bento/blob/250efaa/scripts/debian/networking.sh#L8

### DIFF
--- a/scripts/ubuntu/networking.sh
+++ b/scripts/ubuntu/networking.sh
@@ -5,7 +5,7 @@
 rm -f /etc/udev/rules.d/70-persistent-net.rules;
 mkdir -p /etc/udev/rules.d/70-persistent-net.rules;
 rm -f /lib/udev/rules.d/75-persistent-net-generator.rules;
-rm -rf /dev/.udev/ /var/lib/dhcp3/*;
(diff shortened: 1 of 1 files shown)


tersmitten:
In Ubuntu 10.04 it used to be `dhcp3`. But that's probably not a problem :-)

legal90:
Oh, is it? Sorry, I didn't know about that. 
If we care about this case, then I can change my PR by adding some regex there

tersmitten:
`ubuntu-10.04-amd64.json` is still in the repository. Let's make sure both cases work :-)

cheeseplus:
@tersmitten +1- although we'll be moving any deprecated platforms into a separate folder and labeling as such, we definitely want things to still work because folks will for sure continue to use EoL platforms

legal90:
@cheeseplus Done. I've kept the previous dir name and just added `/var/lib/dhcp/*`

cheeseplus:
lgtm

PR #502, opened 2015-12-07
Removed spaces in variable assignment

oskarpearson:
As this is a shell script, we can't have spaces between the variable and it's value


cheeseplus:
Grabbed this in 3c40b4f but thanks for the PR!

PR #508, opened 2015-12-15
Enterprise Linux 7.2 (RHEL/CentOS/OEL)

rickard-von-essen:
Updated RHEL/CentOS/OEL to **7.2**

_Note:_ To install Parallels Tools package ```selinux-policy-devel``` is needed.

All builds passes on Parallels Desktop Pro 11.1.1, VMware Fusion Pro 7.1.3, and VirtualBox 5.0.10. Beside that I haven't done much testing of the boxes.


rickard-von-essen:
When I tested I reverted a8f5ab4b to use ```--default``` instead of ```--force-install```. I think this is because I don't have VMware Fusion 8.

cheeseplus:
+1

PR #510, opened 2015-12-18
Stops DKMS package from being removed.

### LINKED ISSUE #509: Debian cleanup script removes C/C++ compiler and therefore breaks DKMS support
Is this intentional?

This line:
https://github.com/chef/bento/blob/52a13b5cb3146e4fcc81f1d71a483e697244c421/scripts/debian/cleanup.sh#L31

And similarly for Ubuntu

https://github.com/chef/bento/blob/52a13b5cb3146e4fcc81f1d71a483e697244c421/scripts/ubuntu/cleanup.sh#L31

This doesn't seem like you'd want this.  Why install DKMS in the preseed in the first place if you don't need it? or is installing DKMS just a faster way to get everything installed that the Guest Additions require?

If it is not the desired intent, I'd like to submit a PR to remove these lines, someone might need to inspect the other distros as well.

RobertDeRose:
DKMS is needed to recompile the guest additions if a new kernel
is installed.  Removing the C/C++ packages also removes the `dkms`
package.  This fixes #509 for Dedian based distros.

### DIFF
--- a/scripts/debian/cleanup.sh
+++ b/scripts/debian/cleanup.sh
@@ -27,9 +27,6 @@ dpkg --list \
     | grep -- '-dev$' \
     | xargs apt-get -y purge;
 
-# Delete compilers and other development tools
-apt-get -y purge cpp gcc g++;
-
 # Delete X11 libraries
 apt-get -y purge libx11-data xauth libxmuu1 libxcb1 libx11-6 libxext6;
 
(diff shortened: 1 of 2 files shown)


cheeseplus:
lgtm and nice catch

RobertDeRose:
cool! thanks

PR #513, opened 2015-12-21
Revert "fix or suppress all shellcheck warnings"

cheeseplus:
Reverts chef/bento#494 - looks like some of the shellcheck stuff breaks builds. Notably changes to `scripts/debian/update.sh`


cheeseplus:
So some of the changes look like they may be bash compatible but not `sh` compatible which makes for :sadpanda:

PR #533, opened 2016-02-12
Initial commit for adding Fedora 14 to Bento.

sysbot:
The purpose of having Fedora 14 was to build/setup image that's compatible with
Arista EOS [1] which is based on Fedora 14.

Older version of Fedora/RHEL default `sudoers` to `requiretty` which break
Vagrant ability to run shell scripts due to unable to sudo, this is now fixed.

Adding required dkms and other packages to build virtualbox guest properly.

[1] https://eos.arista.com/


cheeseplus:
Mixed feelings here as this is a crazy EOL platform - does Arista not have a spin or something that is public/more up to date?

cheeseplus:
@sysbot maybe label it as Arista EOS then - just don't want to give folks the impression it's something supported. We'll be adding a "deprecated/EOL" folder soonish so it could also live there.

cheeseplus:
Closing this one as it's not something we can reasonably maintain in the pipeline at this time.

PR #541, opened 2016-02-27
Retire deprecated boxes

tas50:
Retire:

Debian 6
Fedora 21
Ubuntu 10.04
Ubuntu 15.04

Also fix the table to include i386 FreeBSD links


cheeseplus:
Something I've meant to do for a bit is not to delete these but move to a deprecated or archive folder - happy to accept the PR but would live to keep the templates, just move label them accordingly

tas50:
Easy enough. I'll move them tomorrow

tas50:
I assume we're ok with not updating the paths to the scripts still work?  Just moving them off to something like "archive" and being done with it.

cheeseplus:
I'd like to write something that can bulk update the paths or make the bento binary do something with it - not critical and we can do as a second pass.

tas50:
Closing this out since we need a better approach

PR #544, opened 2016-03-15
Update mirror URL for Debian 8.2

legal90:
Debian 8.2 has been moved to the "archive", so the default URL should be updated


sethvargo:
LGTM

cheeseplus:
+1

PR #548, opened 2016-03-20
add fake vagrant server

dmlb2000:
This does the minimum to fake a vagrant server off of the builds
generated by bento. This runs out of the root of the bento tree
and uses wsgi and python.

I'm sure you guys would want something more ruby and less python
but I'm only looking for comments and suggestions at this point.

The goal of this is to provide a web service that you can set
VAGRANT_SERVER_URL to that would fake it enough to pull the
boxes from the web service when adding the boxes you build to
vagrant.

### DIFF
--- a/contrib/server.py
+++ b/contrib/server.py
@@ -1,0 +1,114 @@
+#!/usr/bin/python
+
+from __future__ import print_function
+import sys, os
+from distutils.version import StrictVersion
+from argparse import ArgumentParser
+from wsgiref.simple_server import make_server
+import re
+from json import loads, dumps
+
+def load_jsons():
+  images = {}
+  for dirpath, dnames, fnames in os.walk("./builds"):
+    for f in fnames:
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @dmlb2000 

Thank you for opening a PR. I'll defer to other maintainers, but my personal opinion is that this is outside the scope of the Bento project and would best be serviced as its own open source project.

tiwilliam:
I agree with @sethvargo, this is out of scope for bento. I would also like to see this as a standalone project.

yzl:
I also agree that this makes sense as a standalone project.

cheeseplus:
Ditto - also I know for a fact @fnichol wrote something very similar (in Ruby) but I don't think it ever made it public in it's own repo

dmlb2000:
Okay this makes sense, this can be closed I'll see if I can make a project (in Ruby) that does something similar...

PR #553, opened 2016-03-29
Build all changes since last passed build.

rhass:
Summary of Change
================
The previous implementation would only build changes since the last commit. However, this is somewhat flawed in that it may omit changes which were pushed in a series of commits. Moreover, failed builds which have subsequent commits to fix issues previously could have yeilded the incorrect buildset. This now queries the buildkite API to determine the last successful build, and the Git changeset ID associated with the build to determine what addtional changes we should look at since the last good build.

This script now has one new required environment variable which is not set by buildkite normally, which is the `BUILDKITE_ACCESS_TOKEN`.

Open Concerns
============
Given `BUILDKITE_ACCESS_TOKEN` is an environment variable it also exposes a potential security concern. I am not familiar enough with buildkite yet to know if users are able to echo this variable to stdout or pipe it into an email as part of a PR. If this is indeed the case, we may need to use command line options if they cannot view stdout or the command execution,  but this too still has security issues. 

Validation
=======
- [ ] Add the `BUILDKITE_ACCESS_TOKEN` environment variable to the bento buildkite job.
- [ ] Change multiple templates in separate commits and push them both in one push.
- [ ] Change one template or related file in one commit and push the change to make sure there is no regressions.

Suggested Reviewers
=================
- @yzl 
- @cheeseplus 
- @schisamo

### DIFF
--- a/buildkite.rb
+++ b/buildkite.rb
@@ -1,24 +1,100 @@
 #!/usr/bin/env ruby
 
-git_output = `git log -m -1 --name-only --pretty=format:%n`
+require 'json'
+require 'net/http'
+require 'uri'
 
-# TODO: decide what, if anything, is rebuilt if bin/bento, buildkite.rb, scripts/common change.
-changed_files = git_output.split("\n").reject { |t| t.empty? || t.include?('gitignore') || t.include?('dummy_metadata') || t.include?('bin/bento') || t.include?('buildkite.rb') || t.include?('scripts/common') || t.include?('.md')}
+BUILDKITE_ACCESS_TOKEN = ENV['BUILDKITE_ACCESS_TOKEN']
+BUILDKITE_PROJECT = ENV['BUILDKITE_PROJECT'] ||= 'bento'
+BUILDKITE_ORGANIZATION = ENV['BUILDKITE_ORGANIZATION_SLUG'] ||= 'chef'
+IGNORED_FILES = %w(
+  gitignore
+  dummy_metadata
+  bin/bento
+  buildkite.rb
+  scripts/common
+  .md
+).freeze
+
+def buildkite_api_uri(args = {})
+  raise Exception.new('Missing project argument') if args[:project].nil?
+  raise Exception.new('Missing endpoint argument') if args[:endpoint].nil?
+
+  URI::HTTPS.build(
+    host:   'api.buildkite.com',
+    path:   "/v2/organizations/#{BUILDKITE_ORGANIZATION}/pipelines/#{args[:project]}/#{args[:endpoint]}",
+    query:  "access_token=#{BUILDKITE_ACCESS_TOKEN}"
+  )
+end
+
+# Returns an array of builds hashes for the environment defined project.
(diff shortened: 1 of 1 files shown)


buildkite.rb
```diff
@@ -1,9 +1,74 @@
...
+
+  if response.code.to_i > 200
+    raise Exception.new("Unexpected response from BuildKite API: #{response.code_type}")
+  else
+    JSON.parse(response.body)
+  end
+end
+
+def last_passed_build_git_hash
+  @git_hash = ''
```
rhass: I probably don't need an instance variable here.

buildkite.rb
```diff
@@ -1,9 +1,75 @@
...
+  # `rev-list commit_x..HEAD` shows the commit hashes for each commit from a
+  # given commit to the current HEAD.
+  # `--objects` shows the actual files which have changed in the series of
+  # commits.
+  # We strip the commit IDs via awk because its cheap and a lazy way to get only
+  # the data back we need.
+
+  objects = `git rev-list #{last_passed_build_git_hash}..HEAD  --objects | awk '{print $2}'`.split.uniq
+  # Only return objects which are files, or files with a path.
+  objects.select { |object| object if File.file?(object) }
```
schisamo: We can just do our rejection of ignored files here:

```ruby
# this should probably be at the top of the file
IGNORED_FILES = %w(
  gitignore
  dummy_metadata
  bin/bento
  buildkite.rb
  scripts/common
  .md
).freeze

objects.select { |object| File.file?(object) && !IGNORED_FILES.include?(object) }
```

buildkite.rb
```diff
@@ -1,9 +1,75 @@
...
+end
 
 # TODO: decide what, if anything, is rebuilt if bin/bento, buildkite.rb, scripts/common change.
-changed_files = git_output.split("\n").reject { |t| t.empty? || t.include?('gitignore') || t.include?('dummy_metadata') || t.include?('bin/bento') || t.include?('buildkite.rb') || t.include?('scripts/common') || t.include?('.md')}
+# This expects the input array to only contain files and paths to files.
+# eg: [".travis.yml", "CHANGELOG.md","README.md", "debian-8.3-amd64.json"]
+changed_files = changed_files_since_last_passed_build { |t| t.empty? || t.include?('gitignore') || t.include?('dummy_metadata') || t.include?('bin/bento') || t.include?('buildkite.rb') || t.include?('scripts/common') || t.include?('.md')}
```
schisamo: Assuming you make the change I suggested in https://github.com/chef/bento/pull/553/files#r57823160 you could just use `changed_files_since_last_passed_build ` in the rest of the file...no need for this extra `changed_files` variable.

buildkite.rb
rhass: Isn't freeze redundant for a constant?

rhass: Nevermind... seems Ruby allows mutating constants. :sob: 

http://josephyi.com/freeze/

PR #557, opened 2016-04-02
Add FreeBSD 10.3

tas50:
Also move the install script for FreeBSD 10 into a major version directory


cheeseplus:
+1

rickard-von-essen:
It's now released. 👍

tas50:
The build on vbox completes without issues on my local box and then fails, with a problem that I think is related to my local box.  This *should* be good to go though.

cheeseplus:
+1

PR #559, opened 2016-04-04
Add support for Debian 8.4

kenhys:
Debian 8.4 has been released.
https://www.debian.org/News/2016/20160402


rickard-von-essen:
:+1: Tested OK on VirtualBox

PR #562, opened 2016-04-05
Simplify swap and clarify the build tools Fedora version

tas50:
There's no reason to manually partition Fedora.  Let's just do the same thing we do on RHEL.  We get the same result.  Also let's remove the Fedora version from the build tools script since it impacts both 22 and 23.


cheeseplus:
Can't argue with simplifying things if it works better +1

PR #565, opened 2016-04-19
Add new omios box for r151018

tas50:


rickard-von-essen:
👍 Built successfully on VirtualBox 5 and VMware Fussion 7.

Issue #554, opened 2016-03-30
VMware: Ubuntu 15.10 networking issues


rmoriz:
Now with the released images on Atlas, I've the issues already mentioned in the #545.

- Vagrant 1.8.1
- test-kitchen 1.5.0
- VMware Fusion 8.1.0

Problem => No network
Reason => https://github.com/chef/bento/pull/545#issuecomment-202988690
Possible Fix => https://github.com/chef/bento/pull/545#issuecomment-203043020

:exclamation: If you're affected, add a comment **which VMWare product and version** you're using! Thanks! :exclamation:

```shell
kitchen setup 1510
-----> Starting Kitchen (v1.5.0)
-----> Creating <default-ubuntu-1510>...
       Bringing machine 'default' up with 'vmware_fusion' provider...
       ==> default: Box 'opscode-ubuntu-15.10' could not be found. Attempting to find and install...
           default: Box Provider: vmware_desktop, vmware_fusion, vmware_workstation
           default: Box Version: >= 0
       ==> default: Box file was not detected as metadata. Adding it directly...
       ==> default: Adding box 'opscode-ubuntu-15.10' (v0) for provider: vmware_desktop, vmware_fusion, vmware_workstation
           default: Downloading: https://opscode-vm-bento.s3.amazonaws.com/vagrant/vmware/opscode_ubuntu-15.10_chef-provisionerless.box
==> default: Successfully added box 'opscode-ubuntu-15.10' (v0) for 'vmware_desktop'!
       ==> default: Cloning VMware VM: 'opscode-ubuntu-15.10'. This can take some time...
       ==> default: Verifying vmnet devices are healthy...
       ==> default: Preparing network adapters...
       ==> default: Starting the VMware VM...
       ==> default: Waiting for machine to boot. This may take a few minutes...
       Timed out while waiting for the machine to boot. This means that
       Vagrant was unable to communicate with the guest machine within
       the configured ("config.vm.boot_timeout" value) time period.
       
       If you look above, you should be able to see the error(s) that
       Vagrant had when attempting to connect to the machine. These errors
       are usually good hints as to what may be wrong.
       
       If you're using a custom box, make sure that networking is properly
       working and you're able to connect to the machine. It is a common
       problem that networking isn't setup properly in these boxes.
       Verify that authentication configurations are also setup properly,
       as well.
       
       If the box appears to be booting properly, you may want to increase
       the timeout ("config.vm.boot_timeout") value.
>>>>>> ------Exception-------
[...]

PR #545, opened 2016-03-15
WIP: Adding Ubuntu 16.04

cheeseplus:
Adding Ubuntu Xenial (16.04) builds using daily images until released. Verified these build at least.

PR #566, opened 2016-04-19
vmware: tie network device to pci slot 32, see #554

rmoriz:
see #554 #545


cheeseplus:
Going to merge this for now because it fixes things but probably won't break much at this point.

PR #569, opened 2016-04-23
Update omnios ISO SHAs

tas50:
OmniOS re-released the ISOs with fixes, but maintained the same release
numbers so the SHAs need to get updated.  These are straight from their
site and confirmed to be valid locally.


cheeseplus:
+1

PR #576, opened 2016-04-27
Possible fix for Fedora 23 vmware issue

### LINKED ISSUE #521: VMware: Fedora 23 box builds but can't connect
Per discussion with @cheeseplus both of us have experienced a problem with the Fedora 23 VMWare Fusion/Workstation box; it can be built from the definitions here, but will never start in Vagrant. Capturing the issue here.

cheeseplus:
Similar to how the latest vmware and vagrant has issues related to the pci slot change, this seems to fix #521


tas50:
Seems reasonable until (if ever) VMware ships a fix

PR #583, opened 2016-04-29
Switching to more reliable mirror

cheeseplus:


tas50:
Works for me as long as it's more stable

cheeseplus:
Yea, the existing one just results in forever hangs - folks can still sub in their own mirror with a variable as well but this is a better default based on my testing.

Issue #556, opened 2016-04-01
VMware: HGFS not working - Ubuntu 14.04 2.2.5


ayanich:
vmware 8.1 + ubuntu 14.04 v2.2.5  - i tested this build and it still hangs on HFGS kernel module
```
==> default: Waiting for HGFS kernel module to load...
The HGFS kernel module was not found on the running virtual machine.
This must be installed for shared folders to work properly. Please
install the VMware tools within the guest and try again. Note that
the VMware tools installation will succeed even if HGFS fails
to properly install. Carefully read the output of the VMware tools
installation to verify the HGFS kernel modules were installed properly.
```

cheeseplus:
Opened this one just a few seconds after I opened one - renamed for clarity.

moonglum:
Yep, Ubuntu 14.04 v2.2.3 is working fine with HGFS. Is there any way to help find the problem?

cheeseplus:
@moonglum what kernel does 2.2.3 report?

This is probably going back to issues with HWE and the way around that is to start off with an older ISO and dist-upgrade.

moonglum:
@cheeseplus Here is the output for `uname -a`

```
Linux vagrant 3.19.0-25-generic #26~14.04.1-Ubuntu SMP Fri Jul 24 21:16:20 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

cheeseplus:
@moonglum so this is what I expected - this is very much related to https://github.com/chef/bento/issues/529. When you grab the latest 14.04 ISO they do you the courtesy of upgrading your kernel for hardware compatibility purposes. In order to get a non-HWE kernel you have to use an older ISO and upgrade from it.

moonglum:
Upgrading with dist-upgrade, right?

cheeseplus:
Correct, the deb-ish distros do dist-upgrade - it's just that if you start with 14.04.3 you'll get HWE even if you don't want it.

cheeseplus:
Correct, the deb-ish distros do dist-upgrade - it's just that if you start with 14.04.2 you'll get HWE even if you don't want it.

https://wiki.ubuntu.com/Kernel/LTSEnablementStack

davidmnoriega:
Just tested the new 2.2.7 release and its working.

cheeseplus:
Closing!

moonglum:
Yeha! Thank you :)

geerlingguy:
[...]

Issue #529, opened 2016-02-01
Ubuntu: Don't use LTS Hardware Enablement Stack


legal90:
Was originally reported by @byronclark here: https://github.com/Parallels/vagrant-parallels/issues/247

> The current Parallels Ubuntu 14.04 boxes use the LTS Hardware Enablement Stack (https://wiki.ubuntu.com/Kernel/LTSEnablementStack).
Usually this wouldn't be a big issue, but there's a fairly nasty bug that affects NFS and the install(1) command: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1445664.
I'd love to see these boxes built without the LTS Hardware Enablement Stack as recommended by Ubuntu for cloud and virtual images.

To Bento maintainers: What do you think about it? Is it reasonable to disable "LTS Hardware Enablement Stack" in bento templates for Ubuntu?

yzl:
I am :+1: with not using the LTS Hardware Enablement Stack going forward.

tas50:
We should probably disable it since the AWS AMIs don't use it. You'd be testing on a significantly different host than what you'd be using in prod if you were on AWS.

cheeseplus:
I'm down with the change

sethvargo:
This seems reasonable to me

cheeseplus:
Seems we're now being bit by this with specifically VMware tools and it appears the only way to disable this is to start from an older image and upgrade. It's less than ideal but it may be the only solution.

cheeseplus:
It seems that the only way to get non-HWE kernels is to start with older releases and upgrade all the way through - this is pretty lame on Canonical's part but I think that is how we best address this unfortunately.

PR #584, opened 2016-04-29
Fix for VMware HGFS on Ubuntu 14.04

davidmnoriega:
This PR is to address #556 and #529 by using 14.04.1 as the base ubuntu image in the build. This is the recommended way to not use the LTS Enablement Stack kernel, see https://wiki.ubuntu.com/Kernel/LTSEnablementStack

The real solution is for vagrant to update the vmware plugin to not expect the HGFS kernel module as that has been moved into userspace as part of open-vm-tools. See mitchellh/vagrant#6775 for background


cheeseplus:
Was going to do this for next release anyway 👍

PR #587, opened 2016-05-05
HWE BEGONE!

cheeseplus:
A PR already addressed getting 14.04 builds to not use HWE and this brings the same for 12.04. Fixes https://github.com/chef/bento/issues/529


tas50:
+1

PR #598, opened 2016-05-16
Fix vmware vmtools by removing errant character

cheeseplus:


cheeseplus:
Merging as it's broken otherwise.

PR #603, opened 2016-05-20
Move all the logic out of Rakefile and into bin/bento

cheeseplus:
* take all the logic in Rakefile and migrate to bento CLI tool
* 1:1 translation at present
  * plan to make test and upload take arguments
* headless now an option, not hardcode based on platform
* miscellaneous updates to support Linux builders
* remove all bento-* vagrant boxes before AND after test runs

### DIFF
--- a/bin/bento
+++ b/bin/bento
@@ -6,12 +6,16 @@ Signal.trap("INT") { exit 1 }
 $stdout.sync = true
 $stderr.sync = true
 
-require "benchmark"
-require "digest"
-require "json"
-require "optparse"
-require "ostruct"
-require "tempfile"
+require 'benchmark'
+require 'digest'
+require 'json'
+require 'optparse'
+require 'ostruct'
+require 'tempfile'
+require 'net/http'
+require 'aws-sdk'
+require 'kitchen'
+require 'mixlib/shellout'
 
 class Options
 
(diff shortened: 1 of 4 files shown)


cheeseplus:
Going to merge this sucker as it's really only relevant to the pipeline and I've been doing a bunch of testing along the way.

bin/bento
```diff
@@ -331,6 +508,296 @@ class NormalizeRunner
...
+  end
+
+  private
+
+  def delete_version(boxname, version)
+    banner("Deleting version #{version} of box #{boxname}")
+    req = request('delete', "#{atlas_api}/box/#{atlas_org}/#{boxname}/version/#{version}", { 'access_token' => atlas_token }, { 'Content-Type' => 'application/json' })
+
+    case req.code
+    when '200'
```
rhass: Codes 200-399 are all valid responses, where 3xx is usually some sort of redirect.

bin/bento
```diff
@@ -331,6 +508,296 @@ class NormalizeRunner
...
+  private
+
+  def release_version(boxname, version)
+    banner("Releasing version #{version} of box #{boxname}")
+    req = request('put', "#{atlas_api}/box/#{atlas_org}/#{boxname}/version/#{version}/release", { 'access_token' => atlas_token }, { 'Content-Type' => 'application/json' })
+    if req.code == '200'
+      banner("Version #{version} of box #{boxname} has been successfully released")
+    else
+      warn("Something went wrong #{req.code}")
+    end
```
rhass: Consider using `req.value` here as it effectively does the same thing.

PR #606, opened 2016-06-02
Fix vagrant.sh failing on Solaris nodes

tas50:
Redefining home to /home/vagrant fails since that doesn't exist on
Solaris


cheeseplus:
+1

PR #611, opened 2016-06-13
FreeBSD: Properly set roots password.

### LINKED ISSUE #610: FreeBSD: Root password not set!
The FreeBSD 9.3 and 10.3 boxes don't have any root password set allowing any user to become root with `su` without a password.

This is because the install scripts ([9.3](https://github.com/chef/bento/blob/master/http/freebsd-9.3/install.sh#L98) and [10](https://github.com/chef/bento/blob/master/http/freebsd-10/install.sh#L98)) are missing `-h 0`.
E.g.
`echo "vagrant" | pw -V /mnt/zroot/etc usermod root`
should be
`echo "vagrant" | pw -V /mnt/zroot/etc usermod root -h 0`

But fixing this causes the script execution to fail since no root password is given.
```json
      "execute_command": "{{.Vars}} su -m root -c 'sh -eux {{.Path}}'",
```

rickard-von-essen:
Closes #610


rickard-von-essen:
Tested on FreeBSD 10.3 amd64

legal90:
LGTM 👍

Issue #609, opened 2016-06-09
ubuntu-16.04: unattended updates locking dpkg


tpetchel:
Test Kitchen fails to converge bento/ubuntu-16.04 version 2.2.7 when multiple CPUs are specified in `.kitchen.yml`. 

The error message is:

> dpkg: error: dpkg status database is locked by another process

### Repro steps

step 1. Create a cookbook. For example:

```
$ chef generate cookbook ubuntu-test
```

step 2. Modify `.kitchen.yml` to specify 2 CPUs.

```
---
driver:
  name: vagrant

provisioner:
  name: chef_zero

platforms:
  - name: ubuntu-16.04
    driver:
      customize:
        cpus: 2

suites:
  - name: default
    run_list:
      - recipe[ubuntu-test::default]
    attributes:
```

step 3. Run `kitchen converge`

### Result

https://gist.github.com/tpetchel/0b38bb72f296e330c4d90d5c002a63b7

### More info

This was reported on Learn Chef: https://learn.chef.io/manage-a-web-app/ubuntu/run-your-web-application-on-a-node/#comment-2717043449.

After seeing the user's `.kitchen.yml`, I was able to repo on an empty cookbook. The issue does not repro when you do not specify the `customize` section. That's not to say that it's strictly caused by running more than 1 CPU, but perhaps the timing with multi CPUs just happens to make it repro more consistently.

mjuarez:
I'm experiencing the same issue however I've manually set the platform's driver to use 1 cpu.

Relevant output:
https://gist.github.com/mjuarez/ac6bfff8b140ad82399ec93549b5faf1

cheeseplus:
Thanks for the report @mjuarez - when we this was initially reported I expected the CPU count was probably unrelated and more an issue of timing which this seems to support.

cblecker:
+1 to @mjuarez. Am seeing the same thing with 16.04, using 2.2.7 of the bento box with 1 CPU (no customize section).

jrb:
This appears to be caused by a combination of an unattended-upgraded package policy set by the Debian/Ubuntu preseed, and a change in how unattended upgrades are handled.

[...]

PR #612, opened 2016-06-14
Disable automated package upgrades on Debian-like boxes

jrb:
Currently unattended upgrades are enabled, causing unexpected behavior when the daily apt upgrade job runs (with cron.daily on sysv/upstart hosts, and at-boot plus semi-randomly twice a day on hosts with systemd).

Setting this value to none disables the upgrade jobs, in favor of package upgrades only happening when explicitly requested. This would resolve issue #609


cheeseplus:
Don't see any reason not to merge this one at present. +1

Ma27:
@cheeseplus when is the next release planned? I'd need this fix for my ubuntu 16.04 basebox.

tersmitten:
@Ma27 Me too

cheeseplus:
This fix is also incomplete, it requires an additional fix as well - see https://github.com/chef/bento/issues/609#issuecomment-227202607

The releases aren't planned and are based on the amount of free time I have as an open source maintainer. I'm sorry but life and work are very busy for me at the moment.

PR #614, opened 2016-06-16
update apt sources to use archive.debian.org for packages

apollocatlin:
I have added a script to set the debian6 sources to point to archive.debian.org instead of mirrors.kernel.org which is now deprecated for debian6


cheeseplus:
Could we get this for i386 as well?

apollocatlin:
2min.

apollocatlin:
@cheeseplus done.

cheeseplus:
+1

apollocatlin:
👍

Issue #388, opened 2015-07-02
opscode-centos-7.1 Virtualbox box not able to NFS mount


andytson:
It seems a change has been done in a build of opscode-centos-7.1 that is causing for some reason, the private network static ip to not be assigned before vagrant mounts nfs. This causes nfs to fail.

Checking the ip address assignment after the failure shows the static ip isn't assigned. An ifdown and ifup of the private network adapter introduces the static IP.

This wasn't an issue in an earlier opscode-centos-7.1 build (if you did one before that, or it was just me manually building the working one before you released it), nor opscode-centos-7.0.

Using vboxsf, the static ip is assigned to the private network adapter after boot.

Steps to recreate

* import opscode-centos-7.1
* vagrant init opscode-centos-7.1
* edit Vagrantfile to add a private network with a static ip of 33.33.33.15, and add a synced folder with NFS
* vagrant up

causes:
```
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
==> default: Configuring and enabling network interfaces...
==> default: Exporting NFS shared folders...
==> default: Preparing to edit /etc/exports. Administrator privileges will be required...
Password:
==> default: Mounting NFS shared folders...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

mount -o 'vers=3,udp' 33.33.33.1:'/Users/andy' /vagrant

Stdout from the command:



Stderr from the command:

mount.nfs: access denied by server while mounting 33.33.33.1:/Users/andy
```

davemac30:
I believe this is similar to a problem I am having with private_network using chef/centos-7.1.

When my machine first boots, it get it's private network interface (enp0s8, in my case) configured as expected:

```
va ssh -c "ip a s enp0s8"
3: enp0s8: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 08:00:27:26:7d:7e brd ff:ff:ff:ff:ff:ff
    inet 192.168.33.10/24 brd 192.168.33.255 scope global enp0s8
       valid_lft forever preferred_lft forever
    inet6 fe80::a00:27ff:fe26:7d7e/64 scope link 
       valid_lft forever preferred_lft forever
```

A few seconds later, however, the i/f's ipv4 config is gone:

```
$ va ssh -c "ip a s enp0s8"
[...]

PR #617, opened 2016-06-20
Fix new-style device naming from Network Manager on RHEL/CentOS 7

legal90:
Fixes https://github.com/chef/bento/issues/388 and https://github.com/Parallels/vagrant-parallels/issues/265
Details about the root of issue: https://github.com/Parallels/vagrant-parallels/issues/265#issuecomment-226960702

The solution has been taken from Boxcutter: https://github.com/boxcutter/centos/commit/be4174ea6e8c053d25f970d524d1801c6b21c55f

It might look not very clear, so any other solutions are welcome. The current implementation does the following (_only on RHEL/CentOS 7_):
- Cleans up the NetworkManager state and settings before the box packaging
- Appends `/etc/rc.d/rc.local` by the script, which fixes names of network devices in NM configuration
- Being executed only on the first boot, the script mentioned above removes itself from the `/etc/rc.d/rc.local`.

I've verified it with `centos-7.2-x86_64.json` template and `parallels` provider

### DIFF
--- a/scripts/centos/cleanup.sh
+++ b/scripts/centos/cleanup.sh
@@ -23,4 +23,36 @@ for ndev in `ls -1 /etc/sysconfig/network-scripts/ifcfg-*`; do
     fi
 done
 
+# new-style network device naming for centos7
+if grep -q -i "release 7" /etc/redhat-release ; then
+  # radio off & remove all interface configration
+  nmcli radio all off
+  /bin/systemctl stop NetworkManager.service
+  for ifcfg in `ls /etc/sysconfig/network-scripts/ifcfg-* |grep -v ifcfg-lo` ; do
+    rm -f $ifcfg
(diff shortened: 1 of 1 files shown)


yankcrime:
👍 For this - have tested using VMware Fusion and it solves the issue as described.

nabheet:
I have validated that this fix resolves the issue that I discovered with vagrant getting stuck on setting hostname while bringing up a CentOS 7.2 VM using the Parallels provider.

Basically, I can confirm that this is true: `I've verified it with centos-7.2-x86_64.json template and parallels provider`.

Hopefully this will be merged soon.

legal90:
@cheeseplus @rickard-von-essen @yzl We need you LGTMs :)

cheeseplus:
+1

PR #618, opened 2016-07-06
Leave existing kernel and headers

hellysmile:
We need to leave those packages, cuz there is no way to get then anymore if kernel version was updated.

### DIFF
--- a/scripts/centos/cleanup.sh
+++ b/scripts/centos/cleanup.sh
@@ -3,9 +3,6 @@
 # should output one of 'redhat' 'centos' 'oraclelinux'
 distro="`rpm -qf --queryformat '%{NAME}' /etc/redhat-release | cut -f 1 -d '-'`" 
 
-# Remove development and kernel source packages
-yum -y remove gcc cpp kernel-devel kernel-headers perl;
-
 if [ "$distro" != 'redhat' ]; then
   yum -y clean all;
 fi


hellysmile:
Also I have a proposal before installing VBoxLinuxAdditions install dkms and leave it handle all kernel versions updates

legal90:
Hi @hellysmile ,
> there is no way to get then anymore if kernel version was updated.

Do you mean kernel updates happening on the provisioning time (in Packer), or on the run time (in Vagrant)?

`cleanup.sh` is running at the end of provisioning step, _after_ installing guest tools and package updates. So, it shouldn't cause any problems with kernel updates during the box buld.

hellysmile:
In vagrant, in runtime.

If You update kernel-* (for example during yum -y update) after reboot there will be no VBoxGuestAdditions.

My proposal leave this headers and devel packages as well and before installing VBoxGuestAdditions install dkms, which will trigger any further kernel updates

rickard-von-essen:
That also requires `gcc` etc. If you want to upgrade the kernel version you need to install all dependencies for rebuilding the VBoxGuestAdditions your self. 

If you do this frequently it is probably better to use packer to rebuild the box with packer.

rickard-von-essen:
-1

legal90:
I agree with @rickard-von-essen. Such things should be installed during the run time, right before updating the kernel and/or guest tools. 
cc: @cheeseplus

cheeseplus:
I'm also with @legal90 and @rickard-von-essen - these are very much designed to be base boxes with any further customization applied by oneself.

hellysmile:
To update kernel user need remove existing guest addition, install kernel-devel kernel-headers, instal epel repo, install dkms, install guest addition, otherwise `yum update` simply destroy vagrant due broken guest additions

cheeseplus:
We install the latest kernel at build time for the boxes we host on Atlas. Typically this is how you'd consume updates and in the interim if you wanted a specific or newer kernel that would be something taken care of by the individual.

legal90:
@hellysmile I believe [vagrant-vbguest](https://github.com/dotless-de/vagrant-vbguest) plugin should help you to handle this.

rickard-von-essen:
Closing, see comments above.

Issue #600, opened 2016-05-18
Use OVT (Open-VM-Tools) for vmware boxes instead of closed source vmtools


svpace:
As per:
- http://blogs.vmware.com/vsphere/2015/09/open-vm-tools-ovt-the-future-of-vmware-tools-for-linux.html
- https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=2073803

davidmnoriega:
The vagrant vmware plugin still depends on the VMware tools installing the kernel module. https://github.com/mitchellh/vagrant/issues/6775

cheeseplus:
Yep, vagrant still needs the drivers to work properly.

cheeseplus:
Looks like we're no longer blocked as of `4.0.9` of the vagrant-fusion https://github.com/mitchellh/vagrant/issues/6775#issuecomment-222838468

davidmnoriega:
I saw a message somewhere about needing to update the build pipeline before the new boxes are pushed. I just upgraded vmware fusion to 8.5 and ubuntu-14.04 failed trying to mount the shared folder. 

Building the box from source, they come up just fine. Is there a way I can help with the build pipeline?

cheeseplus:
8.5 release came well after the last release and by the looks of it was announced only 2 weeks ago. There isn't anything that anyone can really do to help the build pipeline short of buying me hardware or rackspace.

davidmnoriega:
What are we talking about? I rely on these boxes, so I would be open to making a donation of some type if there was a system in place.

cheeseplus:
Looks like using OVT with CentOS 7 doesn't work for HGFS as you still need to install drivers.

cheeseplus:
We're using OVT in as much as it works, so just ubuntu 16.04 as centos and other ubuntus (w/o HWE don't work). Closing this as there doesn't seem to be any more OVT stuff we can do.

PR #621, opened 2016-07-09
16.04 should use open-vm-tools

cheeseplus:
Eventually will do this for more platforms as makes sense but as 16.04 is broken we'll start there.


cheeseplus:
Partially addresses https://github.com/chef/bento/issues/600

Issue #631, opened 2016-07-26
Add box for openSUSE Leap 42.1 with virtualbox provider to Atlas


MrWong99:
I wasn't able to any find openSUSE Leap 42.1 boxes build for the virtualbox provider. If it can be build please upload it to Atlas.

When I try to build the box by hand it fails during the OS autoinstallation...

cheeseplus:
Last I tried it was failing on all providers but if it's working I am happy to push it.

MrWong99:
The vmware_desktop box is already [available on Atlas](https://atlas.hashicorp.com/bento/boxes/opensuse-leap-42.1/) which is why I thought the box can be build in general.

cheeseplus:
Right, that is the only platform we can get it to build properly on at present.

cheeseplus:
Done - https://atlas.hashicorp.com/bento/boxes/opensuse-leap-42.1

PR #632, opened 2016-07-26
OpenSUSE Leap 42.1: requires 768 Mb memory.

rickard-von-essen:
This seams to improve the reliability of the builds of OpenSUSE Leap 42.1 on all builders.


rickard-von-essen:
Adresses #631

cheeseplus:
+1 - I've been thinking about bumping that up to 1024 across the board even.

PR #638, opened 2016-07-28
Refactored vmware tools scripts

svpace:
Update Centos 7 to use OVT and refactored the scripts structure to make it easier to include new distributions. #600 #621


cheeseplus:
I don't see `scripts/centos/vmware.sh` which was added being called in the PR.

scripts/ubuntu/vmware.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -eux
+
+case "$PACKER_BUILDER_TYPE" in
+vmware-iso|vmware-vmx)
+    apt-get install -y open-vm-tools;
+    mkdir /mnt/hgfs;
+	echo "platform specific vmware.sh executed";
```
cheeseplus: looks like this is indented unnecessarily

svpace: Indeed. Fixed it.

scripts/centos/vmware.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -eux
```
cheeseplus: This file doesn't seem to be getting called by any of the RHEL/CentOS boxes

svpace: Hi @cheeseplus,

Changed it for Centos 7. Still need to test OVT with older versions. What do you mean by "abstract all of them"?

cheeseplus: You've broken out a `vmware.sh` but parallels and virtualbox still share a `vmtools.sh` - it's not necessary for this PR but it feels like we should break them each out or put them all on in one like vmtools.sh.

svpace: Good idea! changed it. See if this is what you have in mind.

Note 1: Although it is possible to select which script to execute in the packer file, it seemed like a lot of json just to get rid of a couple of "cases" so a left them on the scripts

Note 2: I don't have access to parallels so I can not test it. The change is pretty simple so I believe that it should work.

cheeseplus:
Now that I've got 2.2.9 out the door I can look at testing this across all the providers.

svpace:
Hi @cheeseplus, did you had the chance to test this one up?

cheeseplus:
Let's get this in and see how we do.

mvermaes:
Are there any specific host requirements to use this? I tried to build a CentOS 7.2 template incorporating this pull request but the guest is unable to access shared folders, eventually timing out with:

```
HGFS was never found on the running virtual machine. This must be
installed either through VMware tools or Open VM Tools. HGFS is
required for shared folders to work properly. Please install either
the VMware tools or Open VM Tools and try again.
```

cheeseplus:
Doesn't look like this helps for CentOS platforms as there is no HGFS distributed with open-vm-tools so that makes 99% of the reason why we'd use it moot :( 

See: https://github.com/chef/bento/issues/667#issue-179815307

svpace:
Indeed, AFAIK HGFS will only be available when the CentOS team upgrades to OVT 10. I don't use the shared folders feature that much since it is not something I would have in production (i'm a rsync guy myself :) but I imagine that I may be the exception in this case. 

Nonetheless, what do you think about a variable to control this behavior? So we can use the variable file to select between  OVT and VMT boxes?

PR #643, opened 2016-08-01
Add script for sles-12-sp1

oven:
Copied from sles-12 (without sp1). 

Builds without errors, but you need to download the ISO first. Then you can build the image with

    packer build -var 'mirror=/Users/oven/Downloads' -only=virtualbox-iso sles-12-sp1-x86_64.json

Note that this version differs from sles-12 (without sp1) in that the firewall is turned off. That's necessary in order to let the host ssh into the guest.


cheeseplus:
Looks good to me, will hopefully have some other folks test it out internally.

PR #625, opened 2016-07-21
Add support for vsphere post-processsor

hh:
I don't expect this to merge, but I wanted others to be able to take a look at this approach to using jq to merge in / override the post proccessor.

We wanted to benefit from bento within our ESX / VSphere environment.

Cheers,
@hh

PR #647, opened 2016-08-18
Add vSphere postprocessor support

hh:
This is currently performed by injection, but I wanted to give it a broader audience to see how we might incorporate it long term.


cheeseplus:
Super interested in this but I think we could have them sans injection if only to make usage easier for new folks. That or I rewrite everything in a DSL cause it's so much JSON.

Either way, want to get this in somehow

cheeseplus:
Closing in favor of the approach in #625 or something based on that for ease of adoption. Thanks!

PR #654, opened 2016-08-29
Renames "ubuntu-server" task -> "server" for Ubuntu 14.04

### LINKED ISSUE #653: Select the server task instead of the ubuntu-server task in the Ubuntu 14.04 preseed config
In https://github.com/freedomofpress/securedrop/issues/1375#issuecomment-242696534 my colleague points out that in Ubuntu 14.04, the `ubuntu-server` task does not exist. In the comment below I suggest that it has been renamed as the `server` task, which I do think is the case. You probably want to confirm this, but I think what you want to be doing is selecting the `server` task in the preseed config instead of the `ubuntu-server` task.

conorsch:
The Ubuntu 14.04 config was using tasksel to add the "ubuntu-server" task, which doesn't exist. Maybe it used to. The "server" task is actually what we want, and pulls in packages for a headless setup, such
as tmux.

The "ubuntu-server" task was essentially a no-op, so it didn't break anything, but it did cause a lot of packages to be omitted. See [full comparison here](https://gist.github.com/conorsch/7123209a86a2c577ba465b80cb05b0ef).

Closes #653.

### DIFF
--- a/http/ubuntu-12.04/preseed.cfg
+++ b/http/ubuntu-12.04/preseed.cfg
@@ -28,4 +28,4 @@ d-i pkgsel/upgrade select full-upgrade
 d-i time/zone string UTC
 d-i user-setup/allow-password-weak boolean true
 d-i user-setup/encrypt-home boolean false
-tasksel tasksel/first multiselect standard, ubuntu-server
+tasksel tasksel/first multiselect standard, server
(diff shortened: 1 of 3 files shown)


conorsch:
Also affects: 

* Ubuntu 12.04
* Ubuntu 15.10
* Ubuntu 16.04 (uses 15.10 preseed file)

So I've patched those preseed files, as well. There are still a few hits for `ubuntu-server` in the repo, but I haven't explicitly tested rebuilding with the new tasks lists, so I haven't patched those. Happy to rebase or squash these changes as desired.

PR #656, opened 2016-09-07
Add script to disable system sleep for Mac OS X

cblecker:
By default, Mac OS X will sleep after a certain time period. This isn't optimal for Vagrant boxes as you can't SSH to the box when it is sleeping. This change disables sleep at box creation.


rickard-von-essen:
👍

cheeseplus:
LGTM

PR #664, opened 2016-09-23
Add support for macOS Sierra 10.12

legal90:
The template has been copied from `macosx-10.11.json` .I've verified it with `parallels-iso` builder - it works fine.

But I'm not sure - should I change `guest_os_type` value for `vmware-iso` and `virtualbox` builders or not. I'd appreciate if someone verifies the template with these builders.

### DIFF
--- a/macosx-10.12.json
+++ b/macosx-10.12.json
@@ -1,0 +1,234 @@
+{
+  "builders": [
+    {
+      "boot_wait": "2s",
+      "disk_size": "{{user `disk_size`}}",
+      "guest_os_type": "darwin12-64",
+      "headless": "{{ user `headless` }}",
+      "iso_checksum": "{{user `iso_checksum`}}",
+      "iso_checksum_type": "{{user `iso_checksum_type`}}",
+      "iso_url": "{{user `iso_url`}}",
+      "output_directory": "packer-{{user `template`}}-vmware",
+      "shutdown_command": "echo 'vagrant'| sudo -S shutdown -h now",
+      "skip_compaction": true,
+      "ssh_password": "vagrant",
+      "ssh_port": 22,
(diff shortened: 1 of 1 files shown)


cheeseplus:
Trying to test this today - got stuck on softwareupdate earlier but hopefully that was just transient.

legal90:
Yes, `softwareupdate` could take a while because of updates installation. Have you managed to repeat the build?

cheeseplus:
@legal90 it never seems to complete at that stage (I've only tried with Parallels builder) and I'm not sure how long I should be waiting before calling it hung - are you giving it tons of resources or anything? what do your build times look like?

legal90:
No, I use a regular MacBook Air, early 2014. 
But yeah, now I can confirm that the build hangs for more than 1h on this step:
```
==> parallels-iso: Provisioning with shell script: scripts/macosx/update.sh
    parallels-iso: + softwareupdate --install --all
    parallels-iso: Software Update Tool
    parallels-iso: Copyright 2002-2015 Apple Inc.
    parallels-iso:
    parallels-iso: Finding available software
    parallels-iso:
    parallels-iso: Downloading macOS Sierra Update
    parallels-iso: Downloading iTunes

```
and there is still no progress. Probably, it makes sense to try to rebuild the source .dmg and remove all redundant stuff like iTunes from there...

kameghamegha:
I'm testing some more right now, but I was able to get the softwareupdate segment working when I made these changes to the updates script:

```
softwareupdate --background --force --verbose;
softwareupdate --install --all --force --verbose;
```

Not sure the first line is required (I have yet to test without it), but it seems to work with the --force and --verbose options being added.

cheeseplus:
Superseded by #715 but as always the contributions are welcome @legal90

PR #684, opened 2016-10-05
Fix SLES builds

tas50:
1) Suse is Oracle style now and requires a token to download the image
2) Fix the vm tools installs


tas50:
Thoughts @cheeseplus?

PR #693, opened 2016-10-12
Install libpam-systemd on systemd-enabled Debian versions.

### LINKED ISSUE #692: Parallels Debian boxes do not detect halt (need to install libpam-systemd)
The issue was originally reported here: https://github.com/mitchellh/vagrant/issues/6207 with a solution being provided within a comment here:
* https://github.com/mitchellh/vagrant/issues/6207#issuecomment-234332985

jfilip:
Resolves #692.


jfilip:
It looks like #684 will resolve the failures causing my build to fail.

Issue #658, opened 2016-09-15
Set UTF-8 locale by default


hauleth:
Currently locale for Ubuntu 16.04 is `en_US` which forces PostgreSQL to use `LATIN1` charset instead of `UTF8`. It would be nice to have UTF-8 locale enabled by default.

cheeseplus:
Would `en_us.utf8` be good enough?

hauleth:
It would be perfect.

fumiyas:
FYI.

* #531721 - bash: glob pattern "[A-Z]" matches "b", "c" .. and "z" on LC_ALL=en_US.UTF-8 - Debian Bug report logs  
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=531721

I love C.UTF-8 that is available on Ubuntu 16.04, but I don't know PostgreSQL supports it or not.

cheeseplus:
Fixed by #702 - if we need to make anymore changes here we can raise a new issue.

scottjacobsen:
Will a new version of this box be released soon? It would be great to have this so I can get postgres going.

cheeseplus:
For some value of soon - I am currently on the road for work without access to to quality bandwidth that is required to build these in a timely fashion. All the hypervisors dropped new versions quite recently so these have to be accounted for as some of the fixes, especially VirtualBox 5.1.10 which was released 2 days ago, are needed for some platforms to even build.

PR #702, opened 2016-10-21
Use UTF-8 locale on Debian / Ubuntu boxes

iamthad:
Addresses #658 but only for those distros. It's not clear to me where locale is set in the other boxes.


iamthad:
The Travis checks appear to have failed due to the issue fixed by #684

iamthad:
I've marked this as an "obvious fix" in the commit message.

PR #703, opened 2016-10-21
Only build on master

tas50:
Avoid double testing on PRs. Also test on 2.2.5 not 2.2.0

Signed-off-by: Tim Smith tsmith@chef.io


tas50:
@cheeseplus Updated to Ruby 2.3.1 and the latest Packer

PR #707, opened 2016-10-25
SLES fixes

tas50:
Disable the firewall on SLES 12 to match SLES 12 SP1
Properly remove just the DVD source in the cleanup script


cheeseplus:
+1

PR #709, opened 2016-10-27
Avoid breaking of chef package resource on fedora

artem-sidorenko:
If you use chef with test-kitchen on bento fedora boxes and invoke the resource `package`, it will fail because of missing `yum` (there is no `dnf` provider yet, https://github.com/chef/chef/issues/3201).

This PR is a workaround for the time till chef gets the `dnf` provider.

See https://github.com/chef/chef/issues/5492 for details and logs


cheeseplus:
This seems totally reasonable and lgtm!

PR #714, opened 2016-11-08
Do not remove perl during cleanup process

jtreminio:
Perl is required for installing/compiling and running too many programs to make it worth removing - users will simply need to install it back.

### DIFF
--- a/scripts/centos/cleanup.sh
+++ b/scripts/centos/cleanup.sh
@@ -4,7 +4,7 @@
 distro="`rpm -qf --queryformat '%{NAME}' /etc/redhat-release | cut -f 1 -d '-'`" 
 
 # Remove development and kernel source packages
-yum -y remove gcc cpp kernel-devel kernel-headers perl;
+yum -y remove gcc cpp kernel-devel kernel-headers;
 
 if [ "$distro" != 'redhat' ]; then
   yum -y clean all;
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
I somewhat disagree, packages not strictly needed by Vagrant or which are not part of minimal installation shouldn't be included. The are easy enough to install. You can run allot of things on Linux without perl.

cheeseplus:
Thing is perl is almost always part of a base install and a variety of packages, probably incorrectly, depend on it being there. It doesn't add enough weight that I'm concerned about it.

rickard-von-essen:
> and a variety of packages, probably incorrectly, depend on it being there.

If that is the case I think there is a reason to keep it in the box, but to do it just because you don't want have to install it shouldn't be enough.

jtreminio:
@rickard-von-essen sorry, yes, I may has misworded it. Perl is required by enough packages in a vanilla OS that it makes sense to keep it around.

cheeseplus:
Totally, in this case it's more - system stuff will unexpectedly break. The impression when it was first added was that it was _only_ needed for build-tools which after chatting with @jtreminio in IRC seems to not be the case.

cheeseplus:
Also I'll probably close this PR and open a new one so we don't have to have @jtreminio jump through the DCO hoop but I'll keep it open until then.

cheeseplus:
Closing for #731

PR #715, opened 2016-11-11
macOS Sierra support

kameghamegha:
Added Sierra packer template (macos-10.12.json) and updated scripts/macosx/updates.sh to support Sierra.

I've tested this on Virtualbox but don't have access to Parallels or VMware.

### DIFF
--- a/macos-10.12.json
+++ b/macos-10.12.json
@@ -1,0 +1,234 @@
+{
+  "builders": [
+    {
+      "boot_wait": "2s",
+      "disk_size": "{{user `disk_size`}}",
+      "guest_os_type": "darwin12-64",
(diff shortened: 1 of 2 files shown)


kameghamegha:
I tested this with a 10.11 host, packer 0.11.0, virtualbox 5.0.26, and vagrant 1.8.1. I was receiving some rsync failures with the newest versions of virtualbox and vagrant when I tried to do kitchen create/vagrant up, so I rolled them back (not sure which one was giving me the errors).

legal90:
Seems like this PR supersedes https://github.com/chef/bento/pull/664, but here is `softwareupdate` issue fixed, so LGTM 👍

Issue #717, opened 2016-11-16
Ubuntu boxes doubled in size from 12.04 to 16.04 


kursusHC:
Hi and thanks for providing awesome boxes. Looking at the size of Ubuntu VB boxes : 

12.04 : 324 MB
14.04 : 513 MB
16.04 : 641 MB

What is the reason for the huge bump in size?

tas50:
So there's certainly some things we can do to minimize the size of the images and I've done a lot of that in #718. I think you'll see some much smaller images in the next release. The images are going to grow though because the distros are growing. Even with the various new trimming methods I added in #718 here's the the numbers:

Centos 6.8: 254.5
Centos 7.2: 398.9
Ubuntu 12.04: 278.7
Ubuntu 14.04: 403.8
Ubuntu 16.04: 481.9

PR #718, opened 2016-11-17
Reduce size of Linux images

tas50:
In response to issue #717 

The kernel I noticed myself. The others are care of boxcutter. Thanks folks:

- On Ubuntu make sure we cleanup kernels version 4.X instead of just 3.X. This frees up 200+ megs on disk and 50 megs on the image
- Cleanup /var/log files. < 100k on disk
- Purge additional packages. 4.5 megs on disk
- Purge docs packages. 3 megs on disk
- Purge docs
- Purge caches
- Whiteout not only /, but /boot

With these changes my original Ubuntu 16.04 image was 682.7MB and the final file is 481.9MB.

### DIFF
--- a/scripts/common/minimize.sh
+++ b/scripts/common/minimize.sh
@@ -4,6 +4,18 @@ case "$PACKER_BUILDER_TYPE" in
   qemu) exit 0 ;;
 esac
 
+# Whiteout root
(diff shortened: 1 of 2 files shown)


cheeseplus:
love it

PR #723, opened 2016-11-19
Updated to OpenSUSE Leap 42.2

rickard-von-essen:
- Had to reduce the number of packages that is cleaned up for OpenSUSE 42.2.

### DIFF
--- a/http/opensuse-leap-42.2/autoinst.xml
+++ b/http/opensuse-leap-42.2/autoinst.xml
@@ -108,7 +108,7 @@
       <dhcp_hostname config:type="boolean">true</dhcp_hostname>
       <dhcp_resolv config:type="boolean">true</dhcp_resolv>
       <domain>vagrantup.com</domain>
-      <hostname>opensuse-leap-42.1-x64</hostname>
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Done

cheeseplus:
Any reason you chose en_us over en_us.utf8? I ask as we just shifted Debian platforms to en_us.utf8 for things like Postgres.

rickard-von-essen:
> Any reason you chose en_us over en_us.utf8?

The installer fails when using `en_us.UTF-8` and gave a list of supported languages. I don't know why this changed, I guess previously it have failed and the installer defaulted to `en_US` or something.

[A quick googling gave me this](http://doc.opensuse.org/documentation/leap/reference/single-html/book.opensuse.reference/#sec.suse.l10n.examples):

    LANG=en_US.UTF-8

    This is the default setting if American English is selected during installation. If you selected another language, that language is enabled but still with UTF-8 as the character encoding.

cheeseplus:
Weird but that answer works for me!

PR #725, opened 2016-11-22
Updated to Fedora 25

rickard-von-essen:
Tested x86_64 and i386 builds on:
- Parallels Desktop for Mac Pro 12.1.1
- VMware Fusion 7.1.3
- VirtualBox 5.1.6

### DIFF
--- a/fedora-25-i386.json
+++ b/fedora-25-i386.json
@@ -7,7 +7,7 @@
       "boot_wait": "10s",
       "disk_size": "{{user `disk_size`}}",
       "guest_additions_path": "VBoxGuestAdditions_{{.Version}}.iso",
-      "guest_os_type": "Fedora_64",
+      "guest_os_type": "Fedora",
       "hard_drive_interface": "sata",
       "headless": "{{ user `headless` }}",
       "http_directory": "http",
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
I would like to drop `yum` support by default by removing the `dnf-yum` package. I think that should be installed by the ones that need it. Vagrant should work without `yum`.

cheeseplus:
Looks like we're removing DNF from all platforms which I don't think should be the case. Can we do this just for 25+?

tas50:
The reason for adding yum was that chef currently doesn't have a DNF provider for packages. That work is happening now, but without this packages won't install

rickard-von-essen:
> The reason for adding yum was that chef currently doesn't have a DNF provider for packages.

So let's keep it for now. 

> Looks like we're removing DNF from all platforms which I don't think should be the case.

Dnf have been the default since Fedora 22. So it should be possible to remove for all supported versions.

rickard-von-essen:
And I haven't dropped `yum` yet. That requires removing `dnf-yum` from the[kickstart](https://github.com/chef/bento/blob/master/http/fedora-latest/ks.cfg#L31)  file.

PR #727, opened 2016-11-26
Some more help to build RHEL boxes

aetile:
Hello, please review my changes and merge if you find it useful.
Regards,


rickard-von-essen:
There is no reason to treat RHEL any different from CentOS. Most of the changes don't add anything.

aetile:
You are right, but remember that CentOS and RedHat don't use the same RPMs, it can be useful to have separate customization scripts. I had to do some extra work, starting from you CentOS scripts and I just wanted to share this to help people overcome the yum repo difficulty. Personally I used it at work to deploy Oracle DBs. Anyway thanks for your code, was of great help to me.

cheeseplus:
Thanks for the contribution - as @rickard-von-essen mentioned we don't duplicate scripts unless absolutely required because we like to keep this as DRY as possible. This also adds some things like nfs/cifs which if we added would want to be consistent across as many platforms as possible. Hardcoding to a local repo seems clever but I'm not sure people building their own would prefer that over legit repositories - ideally this would be something the user could plug in via their own script. 

For the reasons outlined above I'm closing the PR but we are still interested in making the RHEL experience better just without some of the implementation detail here.

PR #735, opened 2016-12-02
Initial support for SLES 12 SP2

mattiasgiese:
Created a new packer definition and autoyast based on the SLES 12 SP1 ones.


cheeseplus:
lgtm - don't have access to SuSe releases myself so can't really test at this point

PR #739, opened 2016-12-13
Updated to CentOS/RHEL 7.3

rickard-von-essen:


markphip:
Looking forward to moving to the CentOS 7.3 box. What is general timeline for posting new releases to Atlas?

cheeseplus:
We're currently testing these but already encountering issues building vmware drivers on 7.3

mvermaes:
@cheesplus I'm sure you guys already know about it (assuming you are having a similar issue to me), but just in case - https://communities.vmware.com/message/2632077 helped me get vmhgfs compiling again with CentOS 7.3. This was built on top of OVT per https://github.com/mvermaes/centos-vmware-tools/tree/update-for-centos-7.3, but I assume would be a similar issue with only the proprietary tools. Options are to grab the newer Tools from VMware web site, or use the patch provided in their forum.

Hope it helps.

cheeseplus:
Fantastic @mvermaes - we haven't really had time to look into it just yet but this should make life a bit easier! Thanks

PR #743, opened 2016-12-19
Update VMware tools to fix CentOS 7.3 build

### LINKED ISSUE #742: VMware: CentOS 7.3 fails to build
As per https://github.com/chef/bento/pull/739#issuecomment-267633593 we need to update the drivers.

cheeseplus:
There isn't an easy way to grab the latest vmtools from the open web, in this case 10.1.0, so we're using Packer's File provisioner to sideload it just for 7.3 vmware-iso.

Fixes #742

### DIFF
--- a/centos-7.3-x86_64.json
+++ b/centos-7.3-x86_64.json
@@ -135,6 +135,12 @@
       "type": "file"
     },
     {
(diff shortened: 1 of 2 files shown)


cheeseplus:
Build failure is because the tarball for the tools isn't part of the repo - I'm ok with this and this will go away once the version is included in a Fusion/Workstation release.

PR #768, opened 2017-03-12
[macos-10.12] Allow Packer to pick up memory user var.

fnichol:
I'm not 100% sure if this is an issue--it might need to be confirmed but I could not finish building this box without this fix…

---

For some reason with Pack 0.12.3, the current `"memory"` user variable
is not being picked up and a default value of `1024` is always being
used, leading to macOS 10.12 not being able to finish its installation.
For some reason, changing the user variable from `"memory"` to almost
anything else (in this case, `"memsize"`) allows the user variable to be
picked up.

Note that this was observed with the `vmware-iso` builder and may be the
only affected one. I verified this behavior by running:

```
env PACKER_LOG=1 bundle exec bento build --only=vmware-iso --debug macos-10.12
```

### DIFF
--- a/macos-10.12.json
+++ b/macos-10.12.json
@@ -26,7 +26,7 @@
         "hpet0.present": "TRUE",
         "ich7m.present": "TRUE",
         "keyboardAndMouseProfile": "macProfile",
-        "memsize": "{{ user `memory` }}",
+        "memsize": "{{ user `memsize` }}",
         "numvcpus": "{{ user `cpus` }}",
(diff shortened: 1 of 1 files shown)


cheeseplus:
This seems really weird but don't see any open issues on packer regarding it yet, will try to verify but should be safe enough for merge. Curious if it happens for other platforms too.

cheeseplus:
Going to just merge this one for now as I trust it enough to merge. I'll be testing OSX boxes on another pass and will verify then.

PR #778, opened 2017-03-18
Update to latest Debian 8.7 box (64bit only)

kenhys:
NOTE: Debian 8.7 32bit box are not published yet.


kenhys:
Close this PR because 2.3.4 was released. Thanks!

PR #787, opened 2017-04-14
Support CentOS 6.9 box images

kenhys:
Kickstart configuration file (centos-6.8/ks.cfg) is reused as is.


cheeseplus:
Thanks a bunch for the PR, we took #788 only because it already had DCO signoff and was almost identical in terms of changes.

PR #788, opened 2017-04-15
Add CentOS 6.9 boxes.

mrala:
* Use `ks.cfg` from CentOS 6.8


rickard-von-essen:
See #787

cheeseplus:
Merging this PR over #787 as we have the DCO signoff here.

PR #792, opened 2017-04-21
Added debian-9-amd64 builds

### LINKED ISSUE #791: Add Debian 9 (strech)
In the upcoming months strech will be released. Providing early (still testing) builds of strech would allow people to test their stuff on strech.

I'm in the process of re-writing some salt and puppet deployments which I would rather want to base on strech then on jessie at this point in time.

andir:
fixes #791


cheeseplus:
Doesn't look like this has been officially released just yet and we'd definitely want to get 32 bit builds as well (assuming they don't drop it a la CentOS)

rmoriz:
how about renaming that file to something like `debian-9.0-pre-amd64.json`?

cheeseplus:
I'd change the name as suggested and drop the edit to builds.yml - that file is for the build system, not really something anybody else needs to update.

PR #795, opened 2017-04-26
Working around 5.1.20 vbox GA regression

cheeseplus:
The guest additions supplied by 5.1.20 are broken as per https://www.virtualbox.org/ticket/16670

Using the `guest_additions_url` allows us to cache the ISO so we don't have to download it dynamically in each packer instance. This is the least ugly, at least operationally, of the possible solutions and will be removed as soon as upstream fixes land.


cheeseplus:
Also cleaned up some Virtualbox ISO removal code since it's done in one place now.

PR #799, opened 2017-05-01
We definitely still need this for 16.04

cheeseplus:


cheeseplus:
I thought I could clean this up because the bug had been closed in the upstream tracker but this is apparently still a needed thing, at the very least for Ubuntu 16.04.

PR #811, opened 2017-06-05
Added Oracle Linux 6.9 boxes.

StuporHero:
This would be my first contribution, and I am a little confused about the "version" user vars.


tas50:
Thanks @StuporHero. This will get pushed up to Atlas the next time we do a Bento release.

PR #817, opened 2017-06-20
Remove Ubuntu 12.04 / 16.10

tas50:
This can be merged once 16.10 goes EoL in July. Removing 12.04 cleans up some conditionals in the various scripts as well.


tas50:
We're good to merge this at this point. There's no need to push up an updated 16.10 box while 16.10 is going EOL. The previous box is fine for testing if anyone is working on their migration.

PR #822, opened 2017-06-23
Hyper-V support for CentOS 7.3

wickedviking:
Due to the way Hyper-V handles virtual switches the name of the virtual switch must be passed as an environment variable. 

Can be obtained via: ```Get-VMSwitch``` in PowerShell. 

Example packer build command

```
packer build --only=hyperv-iso -var 'hyperv_switch="Your vSwitch Name"' .\centos-7.3-x86_
64.json
```

Several changes were required for the ks.cfg
* ```network --bootproto=dhcp --onboot=on --device=eth0``` forces the network to be available for %post
* ```reboot --eject``` forces ejection of the install media after first reboot, otherwise you end up in the installer again

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,25 +10,25 @@ The following boxes are built from this repository's templates for publicly avai
 
 ### 64 bit
 
-|                    | VirtualBox                     | VMware                            | Parallels                     |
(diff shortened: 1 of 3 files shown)


wickedviking:
bah
Branches!!!

PR #823, opened 2017-06-23
Hyper-V Support for CentOS 7.3

wickedviking:
Due to the way Hyper-V handles virtual switches the name of the virtual switch must be passed as an environment variable. 

Can be obtained via: ```Get-VMSwitch``` in PowerShell. 

Example packer build command

```
packer build --only=hyperv-iso -var 'hyperv_switch="Your vSwitch Name"' .\centos-7.3-x86_
64.json
```

Several changes were required for the ks.cfg
* ```network --bootproto=dhcp --onboot=on --device=eth0``` forces the network to be available for %post
* ```reboot --eject``` forces ejection of the install media after first reboot, otherwise you end up in the installer again

### DIFF
--- a/centos-7.3-x86_64.json
+++ b/centos-7.3-x86_64.json
@@ -103,6 +103,34 @@
     },
     {
       "boot_command": [
+        "<wait5><tab> text ks=hd:fd0:/ks.cfg<enter><wait5><esc>"
+      ],
+      "boot_wait": "10s",
+      "cpu": "{{ user `cpus` }}",
+      "disk_size": "{{user `disk_size`}}",
+      "floppy_files": [
+        "http/centos-7.3/ks.cfg"
+      ],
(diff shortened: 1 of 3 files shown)


centos-7.3-x86_64.json
```diff
@@ -165,6 +193,10 @@
     "headless": "",
     "http_proxy": "{{env `http_proxy`}}",
     "https_proxy": "{{env `https_proxy`}}",
+    "hyperv_generation": "1" ,
+    "hyperv_guest_additions_mode": "disable",
+    "hyperv_guest_additions_path": "https://download.microsoft.com/download/6/8/F/68FE11B8-FAA4-4F8D-8C7D-74DA7F2CFC8C/LinuxIC-4.2.1.iso",
```
rickard-von-essen: Why first disable and then giving a real URL?

PR #830, opened 2017-06-29
Drop 100 megs from the Ubuntu images

tas50:
linux-firmware is the killer fix here, but we get maybe another meg from the others


rickard-von-essen:
:+1:

Issue #711, opened 2016-10-31
ubuntu boxes can fail to upgrade sudo packages noninteratively


smammy:
For Ubuntu boxes, Bento modifies the distribution-provided `/etc/sudoders` file to add `Defaults exempt_group=sudo` and to add `NOPASSWD` to the entry for `%sudo`. This is done in `scripts/ubuntu/sudoers.sh`.

My provisioning process runs `apt-get -qy update && apt-get -qy dist-upgrade` to get latest Ubuntu updates. On my current `bento/ubuntu-14.04-i386` box (v2.3.0), this upgrades the sudo package from 1.8.9p5-1ubuntu1.2 to 1.8.9p5-1ubuntu1.3, which includes a new version of `/etc/sudoers`. The upgrade fails because there is a conffile conflict to resove and the provisioner is non-interactive. Here is the relevant part of the provisioner output:

```
==> default: Setting up sudo (1.8.9p5-1ubuntu1.3) ...
==> default: Configuration file '/etc/sudoers'
==> default:  ==> Modified (by you or by a script) since installation.
==> default:  ==> Package distributor has shipped an updated version.
==> default:    What would you like to do about it ?  Your options are:
==> default:     Y or I  : install the package maintainer's version
==> default:     N or O  : keep your currently-installed version
==> default:       D     : show the differences between the versions
==> default:       Z     : start a shell to examine the situation
==> default:  The default action is to keep your current version.
==> default: *** sudoers (Y/I/N/O/D/Z) [default=N] ?
==> default: Configuration file '/etc/sudoers'
==> default:  ==> Modified (by you or by a script) since installation.
==> default:  ==> Package distributor has shipped an updated version.
==> default:    What would you like to do about it ?  Your options are:
==> default:     Y or I  : install the package maintainer's version
==> default:     N or O  : keep your currently-installed version
==> default:       D     : show the differences between the versions
==> default:       Z     : start a shell to examine the situation
==> default:  The default action is to keep your current version.
==> default: *** sudoers (Y/I/N/O/D/Z) [default=N] ?
==> default: dpkg: error processing package sudo (--configure):
==> default:  EOF on stdin at conffile prompt
```

I can't think of a graceful way to handle this on my end. Forcing 'Y' to this question would clobber bento's changes (which are helpful) and forcing 'N' could break things. (Not in this case, but you get the idea.)

[...]

PR #833, opened 2017-06-30
Move Ubuntu user settings into sudoers.d

brianknight10:
This pull request addresses issue #711.

It moves the settings for the `vagrant` user out of `/etc/sudoers` and into a file under `/etc/sudoers.d`.

Incidentally, this is the practice currently followed by the Debian bento boxes, so this change brings Ubuntu into line.

Thanks for your consideration.

Signed-off-by: Brian Knight <brianknight10@gmail.com>

### DIFF
--- a/scripts/ubuntu/sudoers.sh
+++ b/scripts/ubuntu/sudoers.sh
@@ -1,4 +1,7 @@
 #!/bin/sh -eux
 
 sed -i -e '/Defaults\s\+env_reset/a Defaults\texempt_group=sudo' /etc/sudoers;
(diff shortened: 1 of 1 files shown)


tas50:
Oh yeah much better. Thanks

rmoriz:
beware of https://github.com/chef-cookbooks/sudo/issues/51 when using that image with chef…

tas50:
I'm fairly certain we can flip that now and I was planning to do so. Outside of FreeBSD it looks like we can use .d folders

PR #838, opened 2017-07-04
Make Ubuntu APT use archive.ubuntu.com rather than the US domain

### LINKED ISSUE #810: Ubuntu 14.04 uses US mirror by default
The Ubuntu 14.04 box refers to the US archive for Ubuntu.

According to https://askubuntu.com/a/104701/963 it is possible to refer to a common archive.ubuntu.com which will then resolve:

---

Open your `source.list` using your favorite text editor, ie

    sudo nano /etc/apt/sources.list

locate `http://us.archive.ubuntu.com/ubuntu` and replace it with `http://archive.ubuntu.com/ubuntu`.
---

Has this been considered?

edmorley:
By default the Ubuntu installer sets the repository hostnames in `/etc/apt/sources.list` to the country-specific mirrors based on the locale chosen, which for this image is `us.archive.ubuntu.com`.

The US mirror is slow for users outside the US, so a better option is to use `archive.ubuntu.com`, which automatically resolves to the closest official country mirror to them. This is also the hostname used in the official Ubuntu Docker images.

Fixes #810.

### DIFF
--- a/http/ubuntu/preseed.cfg
+++ b/http/ubuntu/preseed.cfg
@@ -5,6 +5,10 @@ d-i clock-setup/utc-auto boolean true
 d-i finish-install/reboot_in_progress note
 d-i grub-installer/only_debian boolean true
 d-i grub-installer/with_other_os boolean true
+d-i mirror/country string manual
+d-i mirror/http/directory string /ubuntu/
+d-i mirror/http/hostname string archive.ubuntu.com
+d-i mirror/http/proxy string
 d-i partman-auto-lvm/guided_size string max
 d-i partman-auto/choose_recipe select atomic
 d-i partman-auto/method string lvm


edmorley:
The Travis run is failing, but it's the same failure seen on master - for which I've filed #839.

cheeseplus:
👍

edmorley:
Many thanks :-)

NoahO:
Is anyone else seeing this as not fixed on 2.3.8? 

I still see eg:

deb http://**us**.archive.ubuntu.com/ubuntu/ xenial main restricted 

in /etc/apt/sources.list

@edmorley @cheeseplus

edmorley:
Yeah that is due to #864 meaning the config file updated here isn't being used by the virtualbox/vmware boxes.

Issue #724, opened 2016-11-22
Update for VirtualBox 5.1.10


cheeseplus:
https://www.virtualbox.org/wiki/Changelog#v10

cheeseplus:
Thus far this fixing things more than breaks them so all good.

PR #843, opened 2017-07-08
WIP adds hyper-v support for centos-6.9

binamov:
Following @wickedviking example in #823 to help complete #724.

This seems to need an external hyperv virtual switch. Packer by default creates an internal one, so I needed to override from CLI/env:

```
packer build -only=hyperv-iso -var hyperv_switch=bento-hyperv .\centos-6.9-x86_64.json
```

Similarly to #823, I had to add `reboot --eject` to kickstart script, otherwise it was a constant install loop.

Unlike #823, adding `network --bootproto=dhcp --onboot=on --device=eth0` did not help force get the network available for %post, so I installed hyperv-daemons in `packages` section of kickstart script. There's probably a way to fix this...

Signed-off-by: Bakh Inamov <b@chef.io>

### DIFF
--- a/centos-6.9-x86_64.json
+++ b/centos-6.9-x86_64.json
@@ -120,6 +120,30 @@
       "ssh_wait_timeout": "10000s",
       "type": "qemu",
       "vm_name": "{{ user `template` }}"
+    },
+    {
+      "boot_command": [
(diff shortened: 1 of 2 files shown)


binamov:
this is a mostly a dupe of #842 , closing

PR #862, opened 2017-08-01
Add instructions for openSUSE 42.3

mattiasgiese:
This pull request adds support for building openSUSE 42.3 images.

### DIFF
--- a/http/opensuse-leap-42.3/autoinst.xml
+++ b/http/opensuse-leap-42.3/autoinst.xml
@@ -1,0 +1,262 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+
+<!-- http://doc.opensuse.org/projects/autoyast/configuration.html -->
+
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <general>
+    <mode>
+      <confirm config:type="boolean">false</confirm>
+      <forceboot config:type="boolean">true</forceboot>
+      <final_reboot config:type="boolean">false</final_reboot>
+    </mode>
+  </general>
+  <report>
+    <messages>
+      <show config:type="boolean">false</show>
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
Which builders did you test this with?

mattiasgiese:
I could only test with virtualbox-iso for now, will try to build with libvirt/qemu today.

tas50:
The existing 42.2 files can go away and we should probably make the folder path for the auto install file just 42 since there's no need to rename it every time openSUSE releases a point release.

http/opensuse-leap-42.3/autoinst.xml
```diff
@@ -0,0 +1,262 @@
...
+  </bootloader>
+
+  <networking>
+    <ipv6 config:type="boolean">false</ipv6>
+    <keep_install_network config:type="boolean">true</keep_install_network>
+    <dns>
+      <dhcp_hostname config:type="boolean">true</dhcp_hostname>
+      <dhcp_resolv config:type="boolean">true</dhcp_resolv>
+      <domain>vagrantup.com</domain>
+      <hostname>opensuse-leap-42.3-x64</hostname>
```
rickard-von-essen: Unfortunately we have the minior version in the hostname here.

Maybe we can provide the hostname in the `boot_command` as a [parameter](https://doc.opensuse.org/projects/autoyast/#idm140518702500384) to `autoinst.xml`.

tas50: Even if we use the 42 for the hostname it will be correct since that's the major.

mattiasgiese: Yep, that works. I will make these changes and open another pull request

mattiasgiese:
I think i'll try to refactor this one a bit because the added scripts are merely workarounds.

tas50:
I've opened #869 to replace this. Thanks for the initial push

PR #863, opened 2017-08-02
Add instructions for RHEL 7.4

mattiasgiese:
This commit adds support for RHEL 7.4 (using the old centos-7.3 kickstart).
I could only test this with virtualbox-iso so far.


mattiasgiese:
Test with qemu was successful as well.

tas50:
@mattiasgiese Can you remove the existing 7.3 definition. We try to keep just the latest of every major release. There's no need for that file at this point.

tas50:
Thanks

johnjelinek:
How did you get this to build? I get:
```
==> virtualbox-iso: Provisioning with shell script: scripts/centos/update.sh
    virtualbox-iso: + yum -y update
    virtualbox-iso: Loaded plugins: product-id, search-disabled-repos, subscription-manager
    virtualbox-iso: There are no enabled repos.
    virtualbox-iso: This system is not registered with an entitlement server. You can use subscription-manager to register.
    virtualbox-iso: Run "yum repolist all" to see the repos you have.
    virtualbox-iso: To enable Red Hat Subscription Management repositories:
    virtualbox-iso: subscription-manager repos --enable <repo>
    virtualbox-iso: To enable custom repositories:
    virtualbox-iso: yum-config-manager --enable <repo>
==> virtualbox-iso: Unregistering and deleting virtual machine...
==> virtualbox-iso: Deleting output directory...
Build 'virtualbox-iso' errored: Script exited with non-zero exit status: 1
```

tas50:
It looks like this failure is probably caused by https://github.com/chef/bento/commit/a587fdc4fd88dcca3382d28f2eb9acaae59bcbb1. We attempted to add updating to our RHEL like images, but without a valid subscription a true RHEL image will fail to update. We should probably remove that update logic on RHEL. For now you can remove the update.sh script from the package definition and it should complete without issue.

PR #844, opened 2017-07-11
Adding Hyper-V for 64bit Ubuntus

cheeseplus:
Adds Hyper-V build configuration for 14.04-17.04.
- generation 2 (only 64 bit)
- default memory 512->1024

Signed-off-by: Seth Thomas <sthomas@chef.io>

PR #865, opened 2017-08-04
Use correct preseed config with virtualbox/vmware Ubuntu 16.04 boxes 

### LINKED ISSUE #864: Ubuntu 16.04 vagrant/vmware boxes are now using the Hyper-v preseed file
I just tried to confirm the fix for #838 in the newly released 2.3.8, however `/etc/apt/sources.list` was unchanged.

I believe the reason is that #844 inadvertently switched not just the new Hyper-V build to using `ubuntu/preseed-hyperv.cfg`, but also the vagrant/vmware boxes to it too:
https://github.com/chef/bento/commit/5efa9c09762c3974ca35b4d90a83578f3ce4b131#diff-f301f105a4ad959494d6bd5d0eec227aL272

That line needs reverting. 

In addition, #844 ideally should have made the same changes to `preseed-hyperv.cfg` when rebasing against master, as were made to the main preseed file in #838, per:
https://github.com/chef/bento/pull/844#discussion_r127682192

edmorley:
**1) Don't use Hyper-V preseed on virtualbox/vmware Ubuntu 16.04 boxes.**
This reverts the change inadvertently made in #844.

**2) Make Hyper-V Ubuntu boxes use archive.ubuntu.com too**
Applies the fix from #838 to the Hyper-V preseed config added in #844.

Fixes #864.

### DIFF
--- a/http/ubuntu/preseed-hyperv.cfg
+++ b/http/ubuntu/preseed-hyperv.cfg
@@ -5,6 +5,10 @@ d-i clock-setup/utc-auto boolean true
 d-i finish-install/reboot_in_progress note
 d-i grub-installer/only_debian boolean true
 d-i grub-installer/with_other_os boolean true
+d-i mirror/country string manual
+d-i mirror/http/directory string /ubuntu/
+d-i mirror/http/hostname string archive.ubuntu.com
+d-i mirror/http/proxy string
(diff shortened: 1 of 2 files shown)


tas50:
Thanks for the fix

PR #866, opened 2017-08-09
Fedora 24 EOL yesterday

rickard-von-essen:


tas50:
It's as if I just merged this change...

Issue #851, opened 2017-07-19
debian-9.0 update-initramfs is broken due to /etc/udev/rules.d/70-persistent-net.rule being a folder


EugenMayer:
due to https://github.com/chef/bento/blob/master/scripts/debian/networking.sh#L6

`/etc/udev/rules.d/70-persistent-net.rule` is a folder, which seem to confuse update-initramfs:

```
update-initramfs: Generating /boot/initrd.img-4.9.0-3-amd64
       cp: -r not specified; omitting directory '/etc/udev/rules.d/70-persistent-net.rules'
```

Seems to be a known limitation https://bugs.launchpad.net/ubuntu/+source/initramfs-tools/+bug/1636922

either use `ln -s /dev/null /etc/udev/rules.d/70-persistent-net.rules or` whatever it is tried here.

EugenMayer:
@cheeseplus any way to move this forward somehow? i cannot really undestand why the folder has been created in the first place, most probably to force udev to not create network udev rules ( MAC mappings ) since this would do harm for bento boxes.

I think, as you know, this becomes a issue with stretch, since we use the old preseed file here, and not opting out using strict nic namings, thus initramfs wants to establish those -and thus want to remove this file.

a) I am not sure if the proper solution here is switching to strict named interfaces and remove this hack allthogther
b) or opting out from the strict naming and symlink /dev/null to /etc/udev/rules.d/70-persistent-net.rules' instead of creating a folder there.

A different smell tells me, that this folder has been wrong for debian since past days, it was just right for ubuntu and has been reused here.

EugenMayer:
talked to the debian ops, yes debian stretch does use predictable nic names for fresh installations, so its very different from debian jessie here

That said, by default, no udev rules would be generated, since its no longer MAC based.

@cheeseplus could you decide, what you want to stick with, using ethX but avoiding udev rules, so having nic0 on eth0, or do you want to go with the new predictable names?

the ops in debian suggest, opt out from the predictable names, so we need to add

`net.ifname=0` to /etc/default/grub `cmdline_linux='net.ifname=0'`

and then use `ln -s /dev/null /etc/udev/rules.d/70-persistent-net.rule` to avoid persisting the rules.

I am happy to provide a PR here, just need some decisions/guidance

jakob223:
[...]

PR #867, opened 2017-08-09
Disable udev hack on Debain Stretch+

jakob223:
Fixes https://github.com/chef/bento/issues/851 (I think).

This disables a [hack](http://6.ptmc.org/164/) that prevents automatic udev rule creation by placing a folder where there should be a file with udev rules in it. The existence of that folder prevents `libudev-dev` from being installed properly in debian stretch.

I *think* it's no longer necessary to apply this hack to prevent rules from being automatically created, according to comments in #851, but someone should corroborate that claim.


tas50:
Seems like a logical fix. Thanks

PR #871, opened 2017-08-15
Fixes #858

### LINKED ISSUE #858: debian-9 repo sources need to be updated
currently by default apt is pointed at httpredir.deb.debian.org which forwards to fastly.  this is broken and deprecated according to the ops in irc.debian.org.  this should be updated to point to "deb.debian.org"

cheeseplus:

### DIFF
--- a/http/debian-7/preseed.cfg
+++ b/http/debian-7/preseed.cfg
@@ -9,7 +9,7 @@ d-i grub-installer/with_other_os boolean true
 d-i keymap select us
 d-i mirror/country string manual
 d-i mirror/http/directory string /debian
-d-i mirror/http/hostname string httpredir.debian.org
+d-i mirror/http/hostname string deb.debian.org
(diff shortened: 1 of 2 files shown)


RobertDeRose:
Tested locally... works as expected for VMWare Fusion Version 8.5.8.  Can we get a merge and some new boxes pushed.  No idea why release 2.3.8 only shows provider for `vmware` and no longer `vmware-desktop` but vagrant is complaining that there is no provider for vmware fusion. Hoping this corrects that issue as well

cheeseplus:
I think the problem has to do with the migration to Vagrant Cloud as I've noticed the boxes show up differently and I believe there were corresponding changes on the Vagrant side as well to account for this - not positive here but all the boxes started showing up as `vmware` after the migration and the tooling (aside from hostname and token) hasn't changed.

PR #891, opened 2017-08-31
Dawn of folders

cheeseplus:
We've long talked about organizing the templates into folders as opposed to leaving them in the root of the repo and this makes good on those plans.

- all templates for a given platform have been moved to respective directories, i.e. `ubuntu`
- this changes means a user must `cd` into the platform directory before running `packer build`
- `scripts` and `http` resources specific to the platforms have also been moved similarly
- scripts that are common across platforms have moved to `_common`
- `http` and `script` resources that are shared by platforms reside with the most common platform, i.e. `rhel` and `oraclelinux` templates point at the respective `centos` folders
- general cleanup of dead or bad code (empty metadata.json removal)
- updated Readme and examples


cheeseplus:
The check is failing because I need to update the bento-ya gem which is in progress but that shouldn't prevent merge at this point.

tas50:
Lets do it

PR #893, opened 2017-09-08
Add support for SLES 12 SP3

mattiasgiese:
Adjusted packer definition as well as paths. Tested with vagrant 1.9.7 and Virtualbox 5.1.26. libvirt/kvm testing will be done later.


tas50:
I didn’t even realize this had shipped. Thanks

PR #899, opened 2017-09-13
Use the zroot/usr/home file system as location for the vagrant home directory

flimberger:
There is a dedicated ZFS filesystem in zroot/usr/home and mounted under ``/usr/home``, but the ``vagrant`` user uses ``/home``.
On FreeBSD, both variants are valid, but since a dedicated dataset exists, it is less bewildering for FreeBSD users to use that as the home directory location.
For compatibility and convenience, there is also a symlink from ``/home`` to ``/usr/home``.

### DIFF
--- a/freebsd/http/freebsd-11/installerconfig
+++ b/freebsd/http/freebsd-11/installerconfig
@@ -42,10 +42,11 @@ EOT
 touch /etc/fstab
 
 # Set up user accounts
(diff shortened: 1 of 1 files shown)


cheeseplus:
LGTM and that makes sense. Thanks for the contribution!

tas50:
Why not

PR #900, opened 2017-09-14
Ubuntu fixes

Kaurin:
### Builder for all ubuntu versions and architectures
* Parameterized memory and CPU
* ~~Removed "\<wait\>" statements from the builder spec~~

### update.sh for ubuntu
*  Made sure we don't require user interaction as this was breaking this far. Upstream package configs will be provisioned.
* Removed the "reboot" command as that was breaking, and the "sleep" that was unecessary

### networking.sh for ubuntu
* Removed unecessary file deletions (see the GRUB_LINUX_CMDLINE comment below)
* Added a config file for eth0:  `/etc/network/interfaces.d/eth0.cfg`
* Modified GRUB_LINUX_CMDLINE to disable [PredictableNetworkInterfaceNames](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/) -- to avoid getting different interface names on different virtualization platforms.

### ubuntu json files
* Removed  unecessary provisioner block from ubuntu 17.04 i386
* ~~Bumped version to 14.04.04~~
* Bumped version to 16.04.03


Kaurin:
I think something might be wrong with the CI setup. It broke at something unrelated

rickard-von-essen:
The memory settings should be `{{ user `memory` }}`. And the `boot_command` should be the same for all builders (all the back spaces where a workaround for a bug in the Ubuntu installer).

Kaurin:
Thank you. I'll revise.

I didn't touch backspaces, just the wait statements.

I also need to make networking consistent. The networking script does a
good job of cleaning up udev rules, but I still get different interface
names.

For example:
* Interface name during packer  work: ens3
* Interface name when booting the box with vagrant+Libvirt : ens5

This made the network not come up because dhcp was configured only for
ens3.

I'll test a few ideas tonight and submit another pr

cheeseplus:
I've seen the ens3 vs ens5 thing with qemu and was going to fix myself just wasn't a high priority. We could dump consistent named interfaces entirely and drop back to `eth0` but that's more work across the board so fixing this up just for qemu is probably the right call.

Kaurin:
I have an idea on how  to fix it across the board (for Ubuntu), but I need
to test later :)

Kaurin:
I think I've found a solution:
* Add a "eth0" config file into `/etc/network/interfaces.d/eth0.cfg`
* Append the GRUB_CMDLINE_LINUX to include `net.ifnames=0 biosdevname=0`

On a VM boot, the primary network interface will pick up "eth0". Unfortunately, we lose [PredictableNetworkInterfaceNames](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/), but I think that might be acceptable for vagrant.

In case it's not acceptable, we can get rid of GRUB_CMDLINE_LINUX modification, and just add a bunch of config files into `/etc/network/interfaces.d/`

cheeseplus:
My biggest concern here is flipping to `eth0` just for one platform - if we want to disregard [PredictableNetworkInterfaceNames](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/) which I am definitely not opposed to then we need to do it across the board. I've been waffling back and forth on ripping it out again and it seems that BoxCutter has dropped it too so that's more fuel to the argument for `eth0`

cheeseplus:
Also having the waits in there is preferable as not everyone has the same CPUs and hardware - using `PACKER_KEY_INTERVAL=10ms` on your system can speed this up from the default `100ms`.

Kaurin:
Roger, will revert the waits.

I'll wait for further input to as the decision on what to do with interface
naming

cheeseplus:
I'll think about it a bit today and get back to you, it's not bad to go with `eth0` just then I'll want to make sure we're consistent so it's a bit more work updating the rest but probably the right call in the long run because myself, like most, think Predictable names are not great especially for VMs.

Kaurin:
Reverted waits.

cheeseplus:
Awesome, I'm going to do some testing and will get back to you this weekend as we'd love to merge this one.

Kaurin:
If you don't mind testing Ubuntu <= 15 as well, so we can remove that nasty block from networking.sh

Kaurin:
A few more changes. Updated comment, or you can look at individual commits.

I've also developed a few testing scripts, but I'm gonna leave that for a separate PR if I can make them flexible enough for this project. Right now, they are only targeted for qemu testing

ubuntu/scripts/networking.sh
```diff
@@ -1,17 +1,23 @@
 #!/bin/sh -eux
 
-ubuntu_version="`lsb_release -r | awk '{print $2}'`";
-major_version="`echo $ubuntu_version | awk -F. '{print $1}'`";
+# ubuntu_version="`lsb_release -r | awk '{print $2}'`";
```
cheeseplus: Let's go ahead and remove all the commented lines entirely.

ubuntu/scripts/update.sh
```diff
@@ -23,6 +24,4 @@ APT::Periodic::Enable "0";
 EOF
 
 # Upgrade all installed packages incl. kernel and kernel headers
-apt-get -y dist-upgrade;
-reboot;
-sleep 60;
+apt-get -y dist-upgrade -o Dpkg::Options::="--force-confnew";
```
cheeseplus: Do we need to do this? there shouldn't be any conflicts this early on since nothing else should be changed.

ubuntu/ubuntu-14.04-amd64.json
```diff
@@ -284,17 +294,16 @@
     "https_proxy": "{{env `https_proxy`}}",
     "hyperv_generation": "2",
     "hyperv_switch": "{{env `hyperv_switch`}}",
-    "iso_checksum": "946a6077af6f5f95a51f82fdc44051c7aa19f9cfc5f737954845a6050543d7c2",
+    "iso_checksum": "07e4bb5569814eab41fafac882ba127893e3ff0bdb7ec931c9b2d040e3e94e7a",
     "iso_checksum_type": "sha256",
-    "iso_name": "ubuntu-14.04.1-server-amd64.iso",
+    "iso_name": "ubuntu-14.04.4-server-amd64.iso",
```
cheeseplus: This is intentionally not bumped because it introduces the HWE kernel which generally isn't recommended for virtualized hosts. We've debated just going with HWE but it'd be a change in behaviour so we'd want to make sure this doesn't cause an issue.

cheeseplus:
Added comments to a bunch of lines but once those are addressed I'm good to merge and we'll figure out if we convert the rest back to pre-Predictable Names, even OpenSUSE changed it back at the distro level.

ubuntu/scripts/update.sh
cheeseplus: Totally agree it's safe, just wanted to make sure you hit something that required it (which makes sense)

ubuntu/ubuntu-14.04-amd64.json
cheeseplus: so if you install from a later version, you get the HWE kernel by _default_ and the only way to prevent it is using old install media - upgrading doesn't apply HWE if installed in this manner

cheeseplus: Relevant upstream link: https://wiki.ubuntu.com/Kernel/LTSEnablementStack#Ubuntu_14.04_LTS_-_Trusty_Tahr

Kaurin:
Reverted to Ubuntu 14.04.01

cheeseplus:
Awesome, thanks for sticking with it this long! We've discussed a bit internally and think we might start using HWE kernels cause the performance is better in so many cases that it's probably safe to say people want it more than not. But we can do that separately and that makes this good to merge.

edmorley:
Hi! :-)

This change broke some of our Vagrant provision script iptable rules, since they were hardcoding `enp0s3`. This was surprising, since this breaking change isn't mentioned in:
https://github.com/chef/bento/blob/master/CHANGELOG.md

I'll rewrite the scripts to dynamically calculate the interface name (so it works with all image versions, since we can't guarantee our contributors will be on the latest version), using one of these approaches:
https://github.com/hashicorp/vagrant/issues/8322#issuecomment-335340433
https://github.com/hashicorp/vagrant/blob/v2.0.1/plugins/guests/linux/cap/network_interfaces.rb#L19

...but I don't suppose you could update the changelog?

PR #905, opened 2017-09-16
Updating all templates for Packer 1.1.0

cheeseplus:
Adds `expect_disconnect` to all script provisioners because we do reboots and don't want that to blow up.


rickard-von-essen:
I suggest that `update.sh` is broken out to it's own provisioner block and that `expect_disconnect` is only applied to that.

cheeseplus:
That makes sense and was something I thought about doing but wanted to get the templates unbroken first, optimized later.

PR #910, opened 2017-10-03
Fix SLES autoinst.xml path.

foonix:
Fixes incorrect XML path caused by cffc78107e71efe747a76ee5911bd9331cea458d.


cheeseplus:
Thanks so much @foonix

PR #913, opened 2017-10-07
Set default HOME_DIR in virtualbox.sh

danielmai:
This is consistent with the other builder scripts. Also, removing the
hardcoded /home/vagrant string allows non-vagrant user images to be
created.


cheeseplus:
Looks good, thanks for the contribution!

PR #914, opened 2017-10-12
Support Debian 9.2 point release

kenhys:
See https://www.debian.org/News/2017/20171007

Signed-off-by: Kentaro Hayashi <hayashi@clear-code.com>


tas50:
Thanks

PR #921, opened 2017-10-24
🚑 Update path to preseed-hyperv.cfg to remove ubuntu/ folder path

svpernova09:
Noticed when trying to build a hyper-v version of the ubuntu 16.04 AMD64 box that the path was wrong for the preseed file. This PR updates the path to where you should no longer get the `Failed to retrieve the preconfigureation file` error during the build process.


cheeseplus:
Looks good, thanks so much!

Issue #657, opened 2016-09-08
Post processor hook


mrjcleaver:
I'd like a post-processor hook to call an optional script after completion

https://chefcommunity.slack.com/archives/general/p1472577941000023

> > I want to bootstrap a VM on ESXi that sets up DHCP, sets node name and includes chef’s bootstrap.
> 
> bento is about setting up a clean image for things like test-kitchen and chef/ansible/salt/shell to build on top of now you could modify/fork to do custom stuff but that is outside the scope of what bento is about
> cheeseplus [1:59 PM]  
> yea, the only thing we can do to make it a little easier is add an optional script so like you feed it an env var for optional script(s) boxcutter does that and it's the only thing we don't really have parity with
> 
> > Better though - what’s the chance that this optional script would get added to Bento?
> 
> so the fact I've mentioned adding it means I'm up for it, just nobody has asked

mrjcleaver:
This hook and wrapper stuff is related to https://github.com/ianchesal/packer-config/issues/21 / https://github.com/aspring/racker/issues/10 - related projects for which I'm pushing to get documented on https://github.com/aspring/racker/issues/10

There's also opportunity and perhaps benefit to consolidate extensions in Packer-config, racker and this hook stuff for Bento.

mrjcleaver:
> Go is a full fledged language, certainly, but Packer is JSON and not actually a DSL on top of Go
> so Packer templates themselves can't inherently be DRY like you'd like them to so you use generators

mgruner:
I would highly appreciate the possibility of specifying a custom script like in boxcutter. Thanks for considering.

arizvisa:
What I've done to accomplish this is by using `jq` to append a json item to the provisioners. Something like the following will take the contents of `your.json` and append it to the list of provisioners.
```
$ jq --slurpfile item your.json '.provisioners += $item' /path/to/template.json >| /path/to/new/template.json
```

Stromweld:
Closing this due to age. Please feel free to check out the new HCL2 re-write and open a new feature request or PR.

PR #931, opened 2017-11-03
Added support to install Configuration Management clients

pkleanthous:
This change is necessary because:

* Every time that we provision a vagrant box via Kitchen CI, we spend some minutes waiting for the
Configuration Management client to be installed on the target machine.

The issue is resolved in this commit by:

* Add a script that installs the relevant client based on two environment variables:

 1.   CM_PROVISIONER | Supported clients "saltminion" & "chefclient"
 2.   CM_VERSION | The client version that you want to be installed. Default "latest"

### DIFF
--- a/_common/cmprovisioners.sh
+++ b/_common/cmprovisioners.sh
@@ -1,0 +1,44 @@
+#!/bin/sh -eux
+
+# Variables
+URL=""
+INSTALL_SCRIPT="install.sh"
+VERSION_CMD=""
+
+saltminion () {
+  URL="https://bootstrap.saltstack.com"
+  VERSION_CMD="-P stable"
+}
+
+chefclient () {
+  URL="https://omnitruck.chef.io/install.sh"
+  VERSION_CMD="-v"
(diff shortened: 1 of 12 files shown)


cheeseplus:
This is something we already did but the problem is that then we need to maintain another cross-section of boxes. Bento is not _just_ for chef hence why we provide vanilla boxes. If you want to maintain your own boxes that build on top of bento you're more than welcome to do so but this is explicitly something we moved _away_ from.

pkleanthous:
Hi @cheeseplus, 

Thanks for your comments. I understand that it's not feasible to maintain all those boxes. 
I'm just putting the script here if anyone else wants to create and maintain its own box with a client installed.

cheeseplus:
The purpose of this repo, by and large, is to support those boxes hence why we're not going to merge this - it's too specific. What I will be doing instead, as has already been requested, is letting folks plug in their own custom script file as per https://github.com/chef/bento/issues/657.

Issue #932, opened 2017-11-04
sles-12-sp3-x86_64 build hangs at "Language 'en_US-8' was not found."


lucasrangit:
![image](https://user-images.githubusercontent.com/701818/32400284-6682d4f0-c0bb-11e7-925c-f96bee456824.png)

This happens twice. I select "Ok" each time and the installation proceeds as expected.

PR #935, opened 2017-11-08
fixing bad language-setting in sles/sles12-sp3-x86_64

cdumke:
fixing bad language-setting in vmware-iso builder of sles/sles12-sp3-x86_64

* this fixes issue #932


tas50:
Thanks for the fix

Issue #936, opened 2017-11-08
wrong guest_os_type in sles/sles-12-sp3-x86_64 builder vmware-iso


cdumke:
The "guest_os_type" in the vmware-iso builder of sles/sles-12-sp3-x86_64 should be "sles12-64" - the current value is "sles11-64".

PR #937, opened 2017-11-08
fixing guest_os_type for sles-12-sp(2|3)-x86_64 setting it to sles12-64

cdumke:
* fixing issue #936
* tested on VMware Workstation 12 Pro (for sles-12-sp3-x86_64)


tas50:
Thanks

PR #940, opened 2017-11-15
Added support for Fedora 27

Defilan:
Signed-off-by: Christopher P. Maher <defilan@gmail.com>

This adds a Packer template for Fedora 27. This has been tested and run with Virtualbox.

### DIFF
--- a/debian/debian-8.10-amd64.json
+++ b/debian/debian-8.10-amd64.json
@@ -22,7 +22,7 @@
       "boot_wait": "10s",
       "disk_size": "{{user `disk_size`}}",
       "guest_additions_path": "VBoxGuestAdditions_{{.Version}}.iso",
-      "guest_os_type": "Debian",
+      "guest_os_type": "Debian_64",
       "hard_drive_interface": "sata",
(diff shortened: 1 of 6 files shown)


tas50:
From what I read this is going to be workstation since they haven't shipped server yet. We should probably hold off until they ship the new component based server ISO

Defilan:
Ah ok, that makes sense. I agree, let's wait until the iso comes out. Doesn't look like it will be too long. Thanks for catching that.

rickard-von-essen:
Fedora 27 Server seems to be available now.

Defilan:
@rickard-von-essen Awesome, I'll update the code and test it. If it looks good, I'll push it up for review this afternoon. Thanks for the heads up :)

alindeman:
:tada:

tas50:
@Defilan I think your rebase went wrong here.

Defilan:
@tas50 Yeah, you're right. Don't know what the heck I did this time, haha. I'll go ahead and resubmit this evening. Should be GTG later tonight.

PR #943, opened 2017-11-28
Fix network persistence on openSUSE

bkonick:
The cleanup scripts for openSUSE were not disabling udev rules similar to how CentOS is set up.  This creates a problem when mutating the box (using `vagrant-mutate`) for use with libvirt.  The box will hang because the network can't initialize properly.

This is my first PR to bento, so if there is a specific format or anything I should tweak just let me know.

### DIFF
--- a/opensuse/scripts/cleanup.sh
+++ b/opensuse/scripts/cleanup.sh
@@ -3,5 +3,9 @@
 
 zypper -n rm -u gcc make kernel-default-devel kernel-devel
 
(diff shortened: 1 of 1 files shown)


tas50:
Here's the Novell support doc for anyone reading in the future https://www.novell.com/support/kb/doc.php?id=3048119

tas50:
Thanks. Congrats on the first contribution

PR #948, opened 2017-12-12
Support Debian 8.10 point release

kenhys:
see https://www.debian.org/News/2017/20171209


tas50:
Thanks for putting together both of these. We'll see if we can get these build and pushed soon

PR #954, opened 2017-12-17
Don't install src while creating box

lwhsu:
This speeds up box creation a lot.  The box creation is using pkg(1)
to install software so doesn't need source files.  It doesn't make much
sense installing all the source files and then deleting them during
cleanup stage.


rickard-von-essen:
Have you verified that this is not needed to install VMware tools?

lwhsu:
Ahhh... I thought it uses `emulators/open-vm-tools`.  I'll check if it is possible to switch.

rickard-von-essen:
That would be great!

lwhsu:
Ref: #957

PR #957, opened 2017-12-19
Switch to open-vm-tools for FreeBSD on VMWare

lwhsu:


lwhsu:
From https://docs.vmware.com/en/VMware-Tools/10.2/rn/vmware-tools-1020-release-notes.html :

FreeBSD support is not available for VMware Tools 10.2.0 and later as it has been discontinued in favor of open-vm-tools.

It's probably better to switch to open-vm-tools and installing via official package.

Please note this PR needs more testing since I don't have commercial VMWare license so I can only test it with VMWare player and trial version of VMWare fusion.

Issue #958, opened 2017-12-20
VirtualBox Additions in _common/virtualbox.sh script


arizvisa:
There's an assumption made in `_common/virtualbox.sh` that the name of the script for installing the guest additions is `VBoxLinuxAdditions.run`. I'm not sure if it's just me, but the majority of instances of virtualbox that I've seen use `VBoxLinuxAdditions-$ARCH` in their ISO. Due to this assumption, the builds for VirtualBox don't support things like folder sharing or other types of guest interactions.

To solve this on Linux guests, one can probably test the result of `uname -m` to determine which one to use, or just be obnoxious and use `VboxLinuxAdditions*.run`, maybe..

arizvisa:
I discovered that the reason why the majority of instances of Vbox I've seen use that run script appears to be due to my usage of debian. If nobody else cares, this issue can be closed. It had just caught me by surprise that none of the built VMs had the guest additions installed.

PR #966, opened 2017-12-31
Added guest_additions_url as a user-variable for VirtualBox builds.

arizvisa:
By exposing `guest_additions_url` as a variable, one can force packer to use the correct guest-additions iso for VirtualBox instead of packer trusting the results of the global settings (which may be distribution-specific). If an empty string is specified as `guest_additions_url`, packer will resort to the previous behavior.

This would allow the user to work around issue #958 or any similar ones without having to modify the templates directly.

### DIFF
--- a/centos/centos-6.9-i386.json
+++ b/centos/centos-6.9-i386.json
@@ -7,6 +7,7 @@
       "boot_wait": "10s",
       "disk_size": "{{user `disk_size`}}",
       "guest_additions_path": "VBoxGuestAdditions_{{.Version}}.iso",
+      "guest_additions_url": "{{user `guest_additions_url`}}",
       "guest_os_type": "RedHat",
       "hard_drive_interface": "sata",
       "headless": "{{ user `headless` }}",
@@ -181,6 +182,7 @@
     }
   ],
   "variables": {
+    "guest_additions_url": "",
     "box_basename": "centos-6.9-i386",
     "build_timestamp": "{{isotime \"20060102150405\"}}",
     "cpus": "1",
--- a/centos/centos-6.9-x86_64.json
+++ b/centos/centos-6.9-x86_64.json
@@ -7,6 +7,7 @@
       "boot_wait": "10s",
       "disk_size": "{{user `disk_size`}}",
       "guest_additions_path": "VBoxGuestAdditions_{{.Version}}.iso",
+      "guest_additions_url": "{{user `guest_additions_url`}}",
       "guest_os_type": "RedHat_64",
       "hard_drive_interface": "sata",
       "headless": "{{ user `headless` }}",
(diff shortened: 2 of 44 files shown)


arizvisa:
Afaict, the addition of this variable doesn't hurt anything as Packer first checks the state of the `guest_additions_url` variable to find the iso, if it's an empty string then it falls back to locating the iso by searching the global configuration using `Driver.Iso()`.

Location of empty string test for `guest_additions_url`: [builder/virtualbox/common/step_download_guest_additions.go:71](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/common/step_download_guest_additions.go#L71)

cheeseplus:
We had to revert this one as it appears to still treat this one as a failure. I feel like trusting packer to do the right thing here is the best course of action for the boxes we build and has rarely been an issue. Perhaps if we could understand what failure you were seeing instead of bending the whole repo to account for this issue we could come up with a better solution.

arizvisa:
What still treats this one as a failure? Do you mean Travis-CI or something else?

Yeah, all this does is add the ability for one to specify an alternate guest additions .iso for VirtualBox via a packer variable. If someone is trying to build boxes on a box that they can't update the local copy of the guest additions iso, this lets them specify the url to point to the one from oracle's website or wherever.

The user-variable was left defaulted as an empty string which should cause packer to act the same as if the variable wasn't specified.

cheeseplus:
Fails as in, running `packer build` against the templates with this PR failed - it seems the entire build will run and then it fails. As such I had to revert it entirely.

arizvisa:
Ohh ok. Weird. I'll look into it. Thanks for the heads up.

arizvisa:
Okay, figured it out. Actually based on an invalid assumption on my part but due to a possibly unintentional bug in packer.

Here's the code in packer that checks if the GuestAdditionsURL is empty or not:
```
	// Use the provided source (URL or file path) or generate it
	url := s.GuestAdditionsURL
	if url != "" {
		s.Ctx.Data = &guestAdditionsUrlTemplate{
			Version: version,
		}

		url, err = interpolate.Render(url, &s.Ctx)
		if err != nil {
			err := fmt.Errorf("Error preparing guest additions url: %s", err)
			state.Put("error", err)
			ui.Error(err.Error())
			return multistep.ActionHalt
		}
```

`s.GuestAdditionsURL` is checked to see if it's an empty string and that's what determines whether it should fall back to the original methodology or not. The issue is that the interpolation of the `guest_additions_url` variable is happening _after_ that test to see if it's an empty-string. So even tho the variables specifies to use an empty string, the url comparison is checking the emptiness against `{{user variable-name}}` instead of the value of that variable which should be empty..

arizvisa:
I'll submit an issue in packer with a PR, and let you guys know if/when they fix it.

arizvisa:
Related issue is located here for tracking : hashicorp/packer#5756

arizvisa:
Fixed by commit hashicorp/packer@ff3cbbf28c21527116a88ae29b8ab0206f68cd38.

PR #967, opened 2017-12-31
Fixed a mis-spelling of `output_directory` inside the FreeBSD templates.

arizvisa:
The value for `output_directory` in the VMware section of the FreeBSD templates (freebsd/*.json) is spelled incorrectly as `-vmare` which just results in an inconsistent default name when building everything. This PR simply updates it to the correct spelling of `-vmware`.


tas50:
Thanks

PR #969, opened 2018-01-02
Revert "Added guest_additions_url as a user-variable for VirtualBox builds."

cheeseplus:
Reverts chef/bento#966 - unfortunately, packer considers this failure


arizvisa:
The issue that caused packer to consider #966 a failure when testing has been fixed with commit hashicorp/packer@ff3cbbf28c21527116a88ae29b8ab0206f68cd38. I'm not sure how their release schedule works though. Should I re-submit PR #966 when the next release of packer comes out?

PR #972, opened 2018-01-05
Update Oracle 7 to 7.4

tas50:


cheeseplus:
The last time I tried this is broke in not fun ways but it could have been the virt tools at the time.

PR #977, opened 2018-01-17
Enable autologin for vagrant user on macOS

legal90:
Scripts are taken from the project https://github.com/boxcutter/macos
Having "autologin" set to "true" (the default behavior), we will get the "vagrant" user's GUI session started automatically on the VM boot.

It might be required for different provisioning purposes. 

For example, if we use such VM as an [Omnibus](https://github.com/chef/omnibus) builder, then `dmg` packager will fail if there is no GUI session running on the machine:

```
[Compressor::DMG] I | 2018-01-17T09:37:00+01:00 | Rendering `/Users/vagrant/.bundle/ruby/2.4.0/omnibus-fff89b2e4b0c/resources/dmg/create_dmg.osascript.erb' to `/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/myproj20180117-464-rvez6y/create_dmg.osascript'
[Compressor::DMG] I | 2018-01-17T09:37:01+01:00 | Packaging time: 15.551s
The following shell command exited with status 1:

$ osascript "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/myproj20180117-464-rvez6y/create_dmg.osascript"


Output:

(nothing)

Error:

_RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/myproj20180117-464-rvez6y/create_dmg.osascript:302:308: execution error: An error of type -10810 has occurred. (-10810)
```

P.s. All suggestions are welcome. Let me know if you want the python script being added to the `.sh`.

### DIFF
--- a/macos/macos-10.12.json
+++ b/macos/macos-10.12.json
@@ -201,8 +201,14 @@
   ],
   "provisioners": [
     {
+      "destination": "/private/tmp/set_kcpassword.py",
+      "source": "scripts/support/set_kcpassword.py",
+      "type": "file"
+    },
+    {
       "environment_vars": [
         "HOME_DIR=/Users/vagrant",
+        "AUTOLOGIN={{user `autologin`}}",
         "http_proxy={{user `http_proxy`}}",
         "https_proxy={{user `https_proxy`}}",
         "no_proxy={{user `no_proxy`}}"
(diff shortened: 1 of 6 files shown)


cheeseplus:
This seems good to me.

PR #981, opened 2018-01-23
Using more reliable debian mirror

cheeseplus:


rmoriz:
aren't they exactly the same? Do you have issues with SRV DNS records in VMWare Fusion?

cheeseplus:
My understanding is that they are different but regardless the failures were for virtualbox builders as well and changing the mirror is what did the trick.

rmoriz:
`httpredir.debian.org has been discontinued because of lack of maintenance and now redirects to http://deb.debian.org `

https://wiki.debian.org/DebianGeoMirror

cheeseplus:
While that is definitely true, it still fixed the build so ¯\_(ツ)_/¯

PR #991, opened 2018-02-06
Disable uploading vbox guest additions ISO, which is installed via pkg

lwhsu:
To reduce box size.


cheeseplus:
This makes sense, when the vmtools.sh was refactored, the bit of shell that deleted it was itself removed.

PR #995, opened 2018-02-13
Use smaller install media

lwhsu:


cheeseplus:
Does this provide any benefit? I ask because we use the DVDs for other distros on purpose to save network calls at runtime as Packer by default caches ISOs. This change is fine but it's verging on overoptimization for the sake of it.

lwhsu:
I would say the benefit is small but still helps, downloading a ~600MB CD ISO saves more time and bandwidth than downloading a ~3GB DVD ISO. FreeBSD packages updated very fast and normally people just grab packages directly from the pkg mirror. This is the behavior of the scripts under freebsd/ do right now. They do not use packages in ISOs currently, and this behavior doesn't like to be changed in the near future.

BTW, other JSON files under freebsd/ are using CD images so this change is more or less trying to improve consistency.

cheeseplus:
No disagreement here, just wanted to make sure I understood the logic 👍

lwhsu:
No problem, I should have explained the change more instead of just throwing a change.

PR #1008, opened 2018-03-11
add new platform: HardenedBSD 11

nusenu:
These new files are based on the files in the freebsd folder.
I added the new platform also to builds.yml.

Tested with:
```packer build -only=virtualbox-iso hardenedbsd-11-amd64.json```

### DIFF
--- a/builds.yml
+++ b/builds.yml
@@ -28,6 +28,7 @@ public:
 - 'opensuse-leap-42.3'
 - 'freebsd-11.1'
 - 'freebsd-10.4'
+- 'hardenedbsd-11'
 
 slugs:
   'centos-7': 'centos-7.4'
--- a/hardenedbsd/hardenedbsd-11-amd64.json
+++ b/hardenedbsd/hardenedbsd-11-amd64.json
@@ -1,0 +1,227 @@
+{
+  "builders": [
+    {
+      "boot_command": [
+        "<esc><wait>",
+        "boot -s<wait>",
+        "<enter><wait>",
+        "<wait10><wait10>",
+        "/bin/sh<enter><wait>",
+        "mdmfs -s 100m md1 /tmp<enter><wait>",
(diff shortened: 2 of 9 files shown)


cheeseplus:
Thanks for the PR!

In terms of getting this merged, we could merge this but getting it built regularly is a slightly different matter. We try to balance a number of common platforms and the various hypervisors but when we have niche platforms like this it adds to our build complexity.

This is to say we try to keep the scope of what bento _supports_ (in terms of boxes we build for Vagrant Cloud) as tightly as possible to what is supported by `kitchen-vagrant` which is based on the most common platforms we see used. I'll give this some testing but I want to set expectations as to getting this built and hosted as we have limited resources to support every N platform

nusenu:
Thanks for being open about what the chances are to get this actually build.
I use it with kitchen-vagrant for integration testing of an ansible role:
https://github.com/nusenu/ansible-relayor/blob/test-kitchen/.kitchen.yml

The second platform that is missing for us is OpenBSD, would you build OpenBSD boxes if we submit a patch?

thanks,
nusenu

cheeseplus:
Pretty much the same statement with regard to OpenBSD - in the past we've typically turned these down because if we _don't_ build them they bitrot due to lack of testing and if we _do_ build for more esoteric platforms they tend to cause more build failures.

That said we now have something resembling a build farm so adding builds isn't as bad as it used to be so we're much more willing to add builds now than prior. Bonus points the more providers you've already tested for - if a provider doesn't work, it's better to not include it at all so we can align expectations of users. The test we do for boxes is basically "can kitchen-vagrant spin this up without any extra non-sense" and do a normal converge. With Free/Hardened/OpenBSD this is also way easier because we don't even try to support Shared Folders which is 90% of the problems we tend to see.

nusenu:
Thank you for merging!
I made a small additional change in #1011

PR #1013, opened 2018-03-20
Add FreeBSD i386 to builds

lwhsu:


cheeseplus:
We very specifically stopped building these due to issues with memory constraints and ZFS on 32bit (which comes with a very hearty warning).

lwhsu:
I see, it does not really hurt nowadays.  I'll add some tunables to try to esae it.

cheeseplus:
Just to be clear too, I have some editorial control over what we build based on need/usage - in the past, there just weren't enough downloads to justify the extra effort for 32bit FreeBSD. The builds.yml is for _us_ to add/remove at our discretion but I probably need to add some comments to it to make that part clear ;)

lwhsu:
Thanks for your clarification.  If we can't have these boxes because of the downloads, then I guess there is no much thing I can do.  I would like to have these boxes because they are really handy for test and development needs.  If having these doesn't require too much effort, please still consider include them.  I am happy to provide any kind of help.

cheeseplus:
I'm happy to try to provide these as we have build hardware that wasn't present before. I'm working on a tier based classification system that way we can clearly set expectations and prioritize what we build while still publishing images for more "esoteric" platforms/hypervisors.

PR #1020, opened 2018-03-26
opensuse: Create a new 'vagrant' group for vagrant user

hwoarang:
openSUSE does not create a group with the same name as the user and this
breaks some assumptions that user files are owned by 'vagrant:vagrant'.
As such, lets create such a group and put the vagrant user in it.


cheeseplus:
Thanks for this one!

PR #1026, opened 2018-04-16
ubuntu: linux-headers and shell lint

konstruktoid:
Hi,
Using `bionic.vm.box = "bento/ubuntu-18.04"` fails since the package `linux-headers-$(uname -r)` doesn't exist.

```
==> bionic: Checking for guest additions in VM...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

apt-get install -y linux-headers-`uname -r` dkms
[...]
Stderr from the command:

E: Package 'linux-headers-4.15.0-12-generic' has no installation candidate
```
```
vagrant@vagrant:~$ sudo apt-get -y install linux-headers-`uname -r`
Reading package lists... Done
Building dependency tree
Reading state information... Done
Package linux-headers-4.15.0-12-generic is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'linux-headers-4.15.0-12-generic' has no installation candidate
vagrant@vagrant:~$ sudo apt-get -y install linux-headers-generic
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  linux-headers-4.15.0-15 linux-headers-4.15.0-15-generic
The following NEW packages will be installed:
  linux-headers-4.15.0-15 linux-headers-4.15.0-15-generic linux-headers-generic
[...]
```

Also included in the PR is some shellscript lint fixes.

### DIFF
--- a/solaris/scripts/solaris-10/vmtools.sh
+++ b/solaris/scripts/solaris-10/vmtools.sh
@@ -19,14 +19,14 @@ if [ -f /home/vagrant/.vbox_version ]; then
     mkdir /tmp/vbox
     VER=$(cat /home/vagrant/.vbox_version)
     mkdir /cdrom
-    VBGADEV=`lofiadm -a /home/vagrant/VBoxGuestAdditions.iso`
+    VBGADEV=$(lofiadm -a /home/vagrant/VBoxGuestAdditions.iso)
     mount -o ro -F hsfs $VBGADEV /cdrom
     pkgadd -a /tmp/nocheck -d /cdrom/VBoxSolarisAdditions.pkg < /tmp/allfiles
     umount /cdrom
     lofiadm -d $VBGADEV
     rm -f /home/vagrant/VBoxGuestAdditions.iso
 else
-    VMTOOLSDEV=`/usr/sbin/lofiadm -a /home/vagrant/solaris.iso`
+    VMTOOLSDEV=$(/usr/sbin/lofiadm -a /home/vagrant/solaris.iso)
     mkdir /cdrom
     mount -o ro -F hsfs $VMTOOLSDEV /cdrom
     mkdir /tmp/vmfusion-archive
--- a/solaris/scripts/vagrant.sh
+++ b/solaris/scripts/vagrant.sh
@@ -4,16 +4,16 @@
 HOME_DIR="${HOME_DIR:-/export/home/vagrant}";
 
 pubkey_url="https://raw.githubusercontent.com/mitchellh/vagrant/master/keys/vagrant.pub";
-mkdir -p $HOME_DIR/.ssh;
+mkdir -p "$HOME_DIR/.ssh";
 if command -v wget >/dev/null 2>&1; then
-    wget --no-check-certificate "$pubkey_url" -O $HOME_DIR/.ssh/authorized_keys;
+    wget --no-check-certificate "$pubkey_url" -O "$HOME_DIR/.ssh/authorized_keys";
 elif command -v curl >/dev/null 2>&1; then
-    curl --insecure --location "$pubkey_url" > $HOME_DIR/.ssh/authorized_keys;
+    curl --insecure --location "$pubkey_url" > "$HOME_DIR/.ssh/authorized_keys";
 elif command -v fetch >/dev/null 2>&1; then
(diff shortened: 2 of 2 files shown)


cheeseplus:
I'm curious if this is just an 18.04 issue and if the change effects the other platforms that use the same preseed. Have you tested this against 14.04->17.10?

konstruktoid:
Tested againt 140.04 -> 18.04 using
```system: Darwin Kernel Version 17.5.0: Mon Mar  5 22:24:32 PST 2018; root:xnu-4570.51.1~1/RELEASE_X86_64
packer: 1.2.2
vagrant: Vagrant 2.0.2
qemu: QEMU emulator version 2.11.1
```

What I found is that by not including `linux-headers-generic` or `linux-headers-$(uname -r)`, all images was successfully build and could be used by `Vagrant`.

cheeseplus:
The problem there is that you're using the qemu driver, it doesn't require any drivers like any of the other providers. Pretty sure that change will break the installs of most of the others.

konstruktoid:
Sorry to say, I only got access to `Vagrant`, `Docker`, `Qemu`, and `VirtualBox`.
Should I revert the `kernel-headers` and just submit the lint?

cheeseplus:
That would be best until we can determine that it doesn't break the other providers - we don't build QEMU images ourselves though we do allow them to be in the definitions so while we're happy to fix it, it cannot be at the cost of the other providers.

If this is unnecessary for QEMU then you can change _just_ the qemu section to point at an alternate preseed.cfg.

konstruktoid:
Would you like me to go through and lint all *.sh?

cheeseplus:
Generally I'd urge caution here because we've had folks try to do that in the past and it breaks things pretty terribly. We intentionally use /bin/sh on some platforms and in certain scripts.

konstruktoid:
I'll give it a try.

cheeseplus:
The other complication is that for Ubuntu links /bin/sh to Dash which is neither Bash nor Sh so changing standards via linting on "common" scripts will potentially break other platforms that don't lie about their actual shell.

konstruktoid:
True, but setting `#!/bin/sh` you're accepting local variations. `dash` on Ubuntu, `/bin/sh` is  (not linked) `bash` on MacOSX and so on.

And there's currently a mix of ticks and `$(..)` in the script as of today; https://github.com/chef/bento/blob/master/solaris/scripts/solaris-10/vmtools.sh#L20-L22

Submitting commits that just replaces legacy ticks with `$(..)`.

cheeseplus:
Sure, it's just that the local variation is actually what it says on the tin for all of them BUT ubuntu and MacOS (the latter we don't even build for others)

konstruktoid:
Just being curious, how do you test the builds?
Ignoring the fact I don't have access to VMWare or Parallel, and is building on MacOSX.
Isn't `packer build` and starting a Vagrant/Qemu instance with the freshly built box enough verification that the script works?

cheeseplus:
I have a vSphere cluster in a datacenter that runs builders for each provisioner type that we _fully_ support - VirtualBox, Parallels, VMware Fusion/Workstation. Meaning we build templates and publish images to https://vagrantcloud.com/bento for images that fall under this designation. We do currently build Hyper-V images but they are still flagged as _experimental_ because I currently have to build them on my own desktop and they really only work with latest Windows 10/2016 and HyperV.

Notably this does not include QEMU and we've never provided boxes for those because vagrant-libvirt is finicky enough in conjunction with all the versions of KVM/kernel that are possible that it's never really been worth it to do so. There is also a bit of an expectation that folks using qemu/kvm/libvirt have a bit deeper knowledge and are capable of building these themselves.

There isn't any CI currently but that's mostly due to the nature of how long the builds take and us only relatively recently having hardware in a datacenter as opposed to building these all at home.

As for the latter question: that's enough verification that the script work on _just_ that platform. The second things like packages start changing that can have dramatic effects on each hypervisor that has to build/install a driver during the packer build. Each hypervisor is it's own unique hell so disturbing the house of cards by fixing/altering a seemingly small thing is very easy to do.

konstruktoid:
Closing.

PR #1030, opened 2018-04-29
fedora, centos, rhel: added deltarpm to kickstart files

muellerbe:
I think this one additional package do not hurt and can be save download size during update


cheeseplus:
I don't see any issues with adding it either and seems like folks with slower net connections would benefit.

PR #1048, opened 2018-05-17
Scientific Linux packer built vagrant box

githubfoam:
Just tested with virtualbox.
https://www.scientificlinux.org/

### DIFF
--- a/scientificlinux/SL-7-x86_64.json
+++ b/scientificlinux/SL-7-x86_64.json
@@ -1,0 +1,93 @@
+{
+  "builders": [
+    {
+      "boot_command": [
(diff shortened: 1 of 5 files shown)


cheeseplus:
Somewhat hesitant adding back this platform given the low usage.

cheeseplus:
Was this closed because you no longer need it? Many of us have been preparing for and attending ChefConf so we're a bit behind on PRs but we don't necessarily have to close this one.

The reason I mentioned what I did earlier is that, statistically speaking, next to nobody is using SL and that's 3x more builds that we need to do. We can always add definitions without building them but generally I hope to get more traction than just a single user (though we will add those most of the time)

PR #1051, opened 2018-05-23
Scientific Linux packer built vagrant box 

githubfoam:
(seperate PR) 
https://www.scientificlinux.org/

### DIFF
--- a/scientificlinux/SL-7-x86_64.json
+++ b/scientificlinux/SL-7-x86_64.json
@@ -1,0 +1,93 @@
+{
+  "builders": [
+    {
+      "boot_command": [
+        "<up><wait><tab> text ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/{{user `ks_path`}}<enter><wait>"
+      ],
+      "boot_wait": "10s",
+      "disk_size": "{{user `disk_size`}}",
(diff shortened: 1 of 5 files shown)


cheeseplus:
I'm ok with provisionally adding this one but we really should include the other providers (vmware, parallels) if possible. In 6 months we'll check the download counts to see if folks are actively using it - curious if it gets any traction.

githubfoam:
I d like to do that.Problem is I do not know how to test vmware and
parallel? I am windows10/ubuntu on my desktops.Is it possible to test
builds on those hosts? Thnx.

rickard-von-essen:
As long as they VM tools works on Parallels/VMware with CentOS this should just work.

githubfoam:
I do not have licensed parallels or vmware on my laptop.I have windows10 on
my laptop.

cheeseplus:
We can build them, you don't necessarily have to test all of them locally. It's reasonable to assume a base of Vbox, Fusion, and Parallels - other providers like qemu and Hyper-V are added on an "as works" basis.

githubfoam:
how do you do that? is it linked to travis on gitlab? kind of pipeline
running on gitlab with packer builds?

cheeseplus:
It's a pipeline that I maintain and is not publicly visible. I'll be pushing these builds as soon as I can get them working on the hardware.

PR #1053, opened 2018-06-01
Initial windows support

tas50:
This PR builds on the work done in Matt Wrock's https://github.com/mwrock/packer-templates. It's been modified to use Chef for almost all the work (minus final cleanup) and most of the things we did in boxcutter-windows have been moved over to the Chef recipe. Using Chef here is really nice not only because the complexity of the powershell/vbscript/batch melts away, but because we can easily test the logic in test kitchen without having to fully spin up a windows install. The feedback loop is pretty great, and after spending days waiting for boxcutter-windows installs to fail I'd never go back.


jakauppila:
Are the Bento builds forthcoming on `vagrantup.com` ? I see @tas50 has some deployed.

Thanks a lot for doing this! We were actually planning to start building some ourselves as most of the others out there are pretty out of date.

cheeseplus:
No, Windows builds right now are hosted by @tas50 while we figure out the build pipelines and make sure there aren't any issues legally.

PR #1059, opened 2018-06-06
Remove EOL Debian 7

tas50:
Debian 7 went EOL.


rickard-von-essen:
I guess the last commit went into the wrong branch.

tas50:
Reverted my windows commit that shouldn't be there

PR #1066, opened 2018-06-25
Use en_US.UTF-8 locale instead of en_US (which uses ISO-8859-1 codeset)

davejagoda:
Currently, this box uses the `en_US` locale. This in turn uses the `ISO-8859-1` codeset:

```
locale: en_US           archive: /usr/lib/locale/locale-archive
-------------------------------------------------------------------------------
    title | English locale for the USA
   source | Free Software Foundation, Inc.
  address | http://www.gnu.org/software/libc/
    email | bug-glibc-locales@gnu.org
 language | American English
territory | United States
 revision | 1.0
     date | 2000-06-24
  codeset | ISO-8859-1
```

Versions for earlier Ubuntu releases (e.g. 16.04) use the `en_US.UTF-8` locale which is more broadly applicable.

I've tested that this builds a working box for Virtualbox, and corresponds well to the 16.04 box.


davejagoda:
Thanks for merging this. I agree with you regarding consistency.

In case it may help someone else searching for a solution, ISO-8859-1 was causing trouble when using pipenv as part of a vagrant build. Here are some of the more enlightening parts of the vagrant output before this change:

```
default: Creating a virtualenv for this project\u2026
default: Using /usr/bin/python3.6m (3.6.5) to create virtualenv\u2026
```

and

```
default: UnicodeEncodeError: 'latin-1' codec can't encode character '\u2026' in position 50: ordinal not in range(256)
```

With the encoding set to UTF-8, the output looks like this:

```
default: Creating a virtualenv for this project…
default: Using /usr/bin/python3.6m (3.6.5) to create virtualenv…
```

PR #1067, opened 2018-06-26
Set correct permissions on /etc/sudoers.d/vagrant

kbpease:
'visudo -c' reports that /etc/sudoers.d/vagrant should have its mode set to 0440, but it is set to 0644.


kbpease:
@cheeseplus - I can only speak intelligently to centos and rhel - it seems like this will fix both of those, since rhel appears to use the same kickstart files as centos does.

Issue #1070, opened 2018-07-03
Still seeing 'rename failed - No space left on device' with opensuse-leap


bkonick:
I saw that #1016 was supposed to resolve the space issue but even with the latest versions of the box I am getting the error below:

```
-----> Converging <default-opensuse-leap-423>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 6.3.1...
       Removing non-cookbook files before transfer
       Preparing validation.pem
       Preparing client.rb
-----> Installing Chef Omnibus (13)
       Downloading https://omnitruck.chef.io/install.sh to file /tmp/install.sh
       Trying wget...
       Download complete.
       suse 42 x86_64
       Getting information for chef stable 13 for suse...
       downloading https://omnitruck.chef.io/stable/chef/metadata?v=13&p=suse&pv=42&m=x86_64
         to file /tmp/install.sh.3886/metadata.txt
       trying wget...
       sha1	9f2b34993131f4333d10b607cbdaaea337fbbdc6
       sha256	c856fddb8a0f6e86487d74d23013b5a002cfbe918266ca94125592580b3c8a68
       url	https://packages.chef.io/files/stable/chef/13.9.4/sles/12/chef-13.9.4-1.sles12.x86_64.rpm
       version	13.9.4
       downloaded metadata file looks valid...
       /tmp/omnibus/cache/chef-13.9.4-1.sles12.x86_64.rpm exists
       Comparing checksum with sha256sum...
       Installing chef 13
       installing with rpm...
       warning: /tmp/omnibus/cache/chef-13.9.4-1.sles12.x86_64.rpm: Header V4 DSA/SHA1 Signature, key ID 83ef826a: NOKEY
       Preparing...                                                            ################################# [100%]
       Updating / installing...
          1:chef-13.9.4-1.sles12                                               (################################# [100%]
       error: unpacking of archive failed on file /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/libyajl2-1.2.0/ext/libyajl2/vendor/yajl/test/parsing/cases/bignums.json: cpio: rename failed - No space left on device
       error: chef-13.9.4-1.sles12.x86_64: install failed
       Installation failed
       Version: 13
       
       Please file a Bug Report at https://github.com/chef/omnitruck/issues/new
       Alternatively, feel free to open a Support Ticket at https://www.chef.io/support/tickets
       More Chef support resources can be found at https://www.chef.io/support
       
[...]

PR #1072, opened 2018-07-05
openSUSE no space issue and SLES network persistence

bkonick:
* Addresses #1070 for openSUSE
* Implements the same fix in SLES as #943 did for openSUSE


tas50:
Thanks again @bkonick

Issue #1062, opened 2018-06-15
Ubuntu 18.04 gets duplicate DHCP address on Parallels


NoahO:
Hey all, 

I'm having an issue on the latest Ubuntu image (ubuntu-18.04- 201806.08.0) where every Vagrant machine I start gets assigned the same DHCP IP address as the previous.. even vagrant destroying it and recreating it gets the same IP (even though it has a different MAC address). 

EG, 
vagrant up 
==> vm1: Waiting for machine to boot. This may take a few minutes...
    vm1: SSH address: 10.211.55.**3**:22

cd to another project, 
vagrant up 
==> vm2: Waiting for machine to boot. This may take a few minutes...
    vm2: SSH address: 10.211.55.**3**:22

I've been going crazy with this issue, have reinstalled parallels, and I'm about to log a support ticket with them, but this issue definitely doesn't happen with bento/ubuntu-16.04, and doesn't seem to happen with manually installed 18.04... so I'm wondering if anyone else can reproduce this?

cheeseplus:
It's the first I've heard but I also have had issues prior getting 18.04 working for Parallels - we'll need some versions (Vagrant, Parallels, etc) to really dig into reproducing this one but if we can get that or a reproduction repository to try out then we can investigate.

NoahO:
Did some testing today. It *could* be in my environment, but I don't know what's left to reinstall. I have reinstalled Vagrant and cleared ~/.vagrant.d. 

Here's my testing:
```
vagrant --version
Vagrant 2.1.1
```
prlctl --version
prlctl version 13.3.1 (43365)
``
`mkdir 1804Test1 1804Test2
cd 1804Test1
vagrant init bento/ubuntu-18.04
A `Vagrantfile` has been placed in this directory. You are now
ready to `vagrant up` your first virtual environment! Please read
the comments in the Vagrantfile as well as documentation on
`vagrantup.com` for more information on using Vagrant.
vagrant up
Bringing machine 'default' up with 'parallels' provider...
==> default: Registering VM image from the base box 'bento/ubuntu-18.04'...
==> default: Cloning new virtual machine...
==> default: Unregistering the box VM image...
==> default: Setting the default configuration for VM...
==> default: Checking if box 'bento/ubuntu-18.04' is up to date...
==> default: Setting the name of the VM: 1804Test1_default_1529479003824_94151
==> default: Preparing network interfaces based on configuration...
    default: Adapter 0: shared
==> default: Clearing any previously set network interfaces...
[...]

PR #1073, opened 2018-07-06
WIP: Blank netplan machine-id (DUID) so Ubuntu machines get unique ID generated on boot.

NoahO:
This fixes the issue in:
https://github.com/chef/bento/issues/1062

I have not done backtesting on versions other than 18.04, but I can't see why it would cause any issues.


rickard-von-essen:
We just need to ensure that it's regenerated on first boot.

PR #1087, opened 2018-07-13
Fix disk space on SLES 12 and bring back SLES 11

bkonick:
* Implements the same disk space fixes that opensuse has
* Bring back SLES 11 SP4 because it is not EOL. General support ends in 2019, long term in 2022 (see [here](https://www.suse.com/lifecycle/))


tas50:
Thanks

PR #1092, opened 2018-07-26
Update for Ubuntu 18.04.1

JBenPiel:


cheeseplus:
We don't change the names for Ubuntu patch releases so this still should be named `18.04`, not `18.04.1`. If you fix that part then we can get this merged.

JBenPiel:
Good catch, I removed that from the PR.

PR #1094, opened 2018-08-03
Windows 2016: Point to the correct Autounattend.xml

johnjelinek:
This fixes an issue where an unattended install was failing.

Signed-off-by: John Jelinek <jjelinek@containerstore.com>


tas50:
@johnjelinek Thanks!

PR #1095, opened 2018-08-09
Disable consistent naming in RHEL/CentOS 7

jkugler:
Disable consistent naming in RHEL/CentOS 7 so we always have an eth0 interface.


jkugler:
I ran this. It builds the image. A `vagrant up` with this box works. eth0 got the IP 10.0.2.15. `vagrant ssh` worked. `curl www.google.com` worked as well.

PR #1124, opened 2018-10-19
Add ubuntu 18.10 support

chenhan1218:
Add ubuntu 18.10 support


NoahO:
@cheeseplus This works for me (built it myself) but the image doesn't appear publicly. Does this mean it has to be added to builds.yml?

Thanks :)

cheeseplus:
It's two parts: 1) it needs to be added to builds.yml 2) a build (builds) needs to be run manually. I'm not really involved much with the project any longer and no longer have access to the build infrastructure but am helping @tas50 as needed during transition.

PR #1131, opened 2018-11-07
update debian provision: purge installation-report in cleanup.sh

### LINKED ISSUE #1129: Is the debian installer log data kept on purpose?
I've been using veewee for years. New to chef/bento. It's an awesome project. I'm definitely moving over to packer and porting all my old definition files.

Just downloaded the latest Debian box, and looked around for a while. I found that `cleanup.sh` script means to "delete any logs that have built up during the install", as stated in the comment, however, `/var/log/installer/` was still there with a size of 15M (those files don't have '.log' suffix).

Is it kept on purpose?

If not, I'd like to write a patch to remove it during cleanup, which can be done by simply purging the `installation-report` package.

UniIsland:
### Description

The `installation-report` package generates installation report, installer related choices and debug infos into `/var/log/installer/`.

It takes about 15M disk space for debian 9.5 builds.

### Issues Resolved

Fixes #1129

### DIFF
--- a/debian/scripts/cleanup.sh
+++ b/debian/scripts/cleanup.sh
@@ -35,6 +35,7 @@ apt-get -y purge ppp pppconfig pppoeconf;
 
 # Delete oddities
 apt-get -y purge popularity-contest;
+apt-get -y purge installation-report;
 
 apt-get -y autoremove;
 apt-get -y clean;


tas50:
Nice catch.

PR #1144, opened 2018-11-20
Amazon Linux 2 definition

tas50:
Amazon provides a disk image and expects you to create an ISO with cloud-init files, which will be read when the disk boots. This automates creating the ISO, creating the VM from the disk image and exporting the OVF file. Next step is consuming that file in a packer template and then we have a vagrant box.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 .bundle/
-iso
+*.iso
 *.box
+*.ovf
+*.vmdk
+*.vdi
 virtualfloppy.vfd
 packer_cache
 packer.log
--- a/amazon/STEP1_build_ovf.sh
+++ b/amazon/STEP1_build_ovf.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+if [ ! -f amazon.vdi ]; then
+  echo There must be a file named amazon.vdi in this directory!
+  echo You can download the vdi file at https://cdn.amazonlinux.com/os-images/latest/
+  exit 1
+fi
+
+echo "Cleaning up old files"
+rm seed.iso
+rm amazon.ovf
+rm *.vmdk
+
+echo "Creating ISO"
+hdiutil makehybrid -o seed.iso -hfs -joliet -iso -default-volume-name cidata seed_iso
+
+VM="AmazonLinuxBento"
+echo Powering off and deleting any existing VMs name $VM
+VBoxManage controlvm $VM poweroff --type gui 2> /dev/null
+vboxmanage unregistervm $VM 2> /dev/null
+sleep 5
+
+echo "Creating the VM"
+# from https://www.perkin.org.uk/posts/create-virtualbox-vm-from-the-command-line.html
+VBoxManage createvm --name $VM --ostype "RedHat_64" --register
+VBoxManage storagectl $VM --name "SATA Controller" --add sata --controller IntelAHCI
+VBoxManage storageattach $VM --storagectl "SATA Controller" --port 0 --device 0 --type hdd --medium amazon.vdi
+VBoxManage storagectl $VM --name "IDE Controller" --add ide
+VBoxManage storageattach $VM --storagectl "IDE Controller" --port 0 --device 0 --type dvddrive --medium seed.iso
+VBoxManage modifyvm $VM --memory 1024
+VBoxManage modifyvm $VM --cpus 2
+VBoxManage modifyvm $VM --audio none
+VBoxManage modifyvm $VM --ioapic on
+sleep 5
+
+echo Sleeping for 100 seconds to let the system boot and cloud-init to run
+VBoxManage startvm $VM --type gui
+sleep 100
+VBoxManage controlvm $VM poweroff --type gui
+VBoxManage storageattach $VM --storagectl "IDE Controller" --port 0 --device 0 --type dvddrive --medium none
+sleep 5
+
+echo Exporting the VM to an OVF file
+vboxmanage export $VM -o amazon.ovf
+sleep 5
+
+echo Deleting the VM
+vboxmanage unregistervm $VM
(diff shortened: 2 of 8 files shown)


johnjelinek:
for me to get packer to successfully SSH commands against the instance, I immediately needed to enable Password login via `/etc/sshd_config` and change the MAC address to map to what VirtualBox set (different from what it was baked with when running the `STEP1` script. Now it complains that a `vagrant` user wasn't created.

johnjelinek:
I can contribute fixes, but did you have a list of things you really wanted to see completed before this could be merged? Like, as it is, I think STEP1 can only be completed on a mac, which is fine for me, but I dunno if you want that shored up for other initial hosts before this could be merged.

cheeseplus:
The list of things a Vagrant base box needs is pretty well outlined here https://www.vagrantup.com/docs/boxes/base.html

johnjelinek:
k, all that stuff is working as-is aside from the manual config stuff I mentioned above.

cheeseplus:
I was editing the last comment but was going to note that for the foreseeable future the builders will be running macOS so while it'd be nice to design for running on Linux, that's gold plating more than a required functionality.

johnjelinek:
ok, well, from what is already in this branch, I already have a vagrant box running and I'm able to do what I need to with it.

johnjelinek:
should be as-is with something like:
```
platforms:
  - name: amazon-2
    driver:
      username: ec2-user
      password: vagrant
```
I can let you know after I wrap up some other stuff.

cheeseplus:
Ideally folks shouldn't have to supply the username or the password as Vagrant expects a `vagrant` user by default with key based access. At least that would be how I'd go about this along the Path of Least Surprise. In the case of amazon linux I'm not sure how critical the presence of, or more importantly run-time usage of the `ec2-user`, ends up being when not actually in AWS.

tas50:
The goal was to have a working system that would be a drop in replacement for some of the other amazon boxes out there in vagrant. This would require a vagrant user to be setup so a user doesn't have to pass creds. If the creation script only works on a Mac I'm totally ok with that for now. If someone wanted to step up later and make it more cross platform I'd consider that a future enhancement.

johnjelinek:
@tas50: I have some changes that should get you to where you want to be. Do you want me to submit my own PR to your branch or do you want to just bring them in yourself?

Feature Summary:
* Increase time to export ovf from 100 to 200 seconds
This change was necessary for me to get the ovf to boot. Without increasing the time, it would export too early and then packer would only boot to grub.

* Create `vagrant` user as Default EC2 User
This leaves the old `ec2-user` in place, but it's unused. You can `sudo su - ec2-user` if you want to become that user from the `vagrant` user.

* Add `runcmd` commands
These commands allow packer to SSH in with the `vagrant` user and password to continue provisioning. Maybe `/etc/cloud/cloud.cfg.d/90_enable_ssh_pwauth.cfg` can be removed as part of `cleanup.sh`?

I'm able to use a vanilla `vagrant up` and `vagrant ssh` to get into the box without any extra configuration.
```
diff --git a/amazon/STEP1_build_ovf.sh b/amazon/STEP1_build_ovf.sh
index 831530d..31d192b 100755
--- a/amazon/STEP1_build_ovf.sh
+++ b/amazon/STEP1_build_ovf.sh
@@ -33,9 +33,9 @@ VBoxManage modifyvm $VM --audio none
 VBoxManage modifyvm $VM --ioapic on
 sleep 5
 
-echo Sleeping for 100 seconds to let the system boot and cloud-init to run
+echo Sleeping for 200 seconds to let the system boot and cloud-init to run
 VBoxManage startvm $VM --type gui
-sleep 100
+sleep 200
 VBoxManage controlvm $VM poweroff --type gui
 VBoxManage storageattach $VM --storagectl "IDE Controller" --port 0 --device 0 --type dvddrive --medium none
 sleep 5
diff --git a/amazon/amazon-2-x86_64.json b/amazon/amazon-2-x86_64.json
index 03425c6..0eba700 100644
--- a/amazon/amazon-2-x86_64.json
+++ b/amazon/amazon-2-x86_64.json
@@ -10,7 +10,7 @@
       "ssh_port": 22,
       "type": "virtualbox-ovf",
       "source_path": "amazon.ovf",
-      "ssh_username": "ec2-user",
+      "ssh_username": "vagrant",
       "ssh_password": "vagrant",
       "vboxmanage": [
         [
diff --git a/amazon/seed_iso/user-data b/amazon/seed_iso/user-data
index 8eeeff4..222eb4e 100644
--- a/amazon/seed_iso/user-data
+++ b/amazon/seed_iso/user-data
@@ -1,11 +1,17 @@
 #cloud-config
 #vim:syntax=yaml
-users:
-# A user by the name `ec2-user` is created in the image by default.
-  - default
-chpasswd:
-  list: |
-    ec2-user:vagrant
+user: vagrant
+password: vagrant
+
+# install packages for vbox guest additions to complete
+packages:
+  - gcc
+  - kernel-devel
+  - kernel-headers
+  - dkms
+  - make
+  - bzip2
+  - perl
 
 # make sure we don't setup the network on every boot
 write_files:
@@ -14,3 +20,10 @@ write_files:
       # Disable network configuration after first boot
       network:
         config: disabled
+
+  - path: /etc/cloud/cloud.cfg.d/90_enable_ssh_pwauth.cfg
+    content: |
+      runcmd:
+        - sed -i "s/^PasswordAuthentication.*/PasswordAuthentication yes/" /etc/ssh/sshd_config
+        - sed -i "/^HWADDR/d" /etc/sysconfig/network-scripts/ifcfg-eth0
+        - systemctl restart network sshd

```

johnjelinek:
@cheeseplus: what should be done to get TravisCI to pass here? It assumes the `amazon.vdi` is already in place.

PR #1149, opened 2018-12-03
Reduce centos locale size

Sorbog:
### Description

Only one language (en_US.utf8) will be installed. This will reduce the box size ~> 50 mb


cheeseplus:
Turns out that `--instLangs` is an invalid option for CentOS 5.11 
screenshot 2018-12-27 17 05 56

PR #1150, opened 2018-12-03
Remove NetworkManager

Sorbog:
### Description

Move removing NetworkManager to kickstart

### DIFF
--- a/centos/http/7/ks.cfg
+++ b/centos/http/7/ks.cfg
@@ -37,6 +37,13 @@ deltarpm
 -fprintd-pam
 -intltool
 
+# remove NetworkManager
+-NetworkManager
+-NetworkManager-libnm
+-NetworkManager-team
+-NetworkManager-tui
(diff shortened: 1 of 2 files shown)


tas50:
CC: @cheeseplus I know you've gone back and forth on the networkmanager mess for a while

cheeseplus:
This seems to be moving this to kickstart, which if it works, I'm all for since the net effect will be the same. I'm +1 assuming it doesn't uncover some weird edge case.

aequitas:
@Sorbog I noticed (in both approaches) that removing NetworkManager leads to an increased startup time for the VM of around 30 seconds because during boot it times out on the systemd unit `NetworkManager-wait-online.service`. Maybe this could also be removed/disabled?

tas50:
I'm going to close this out at this point since we never heard back about the increased startup time and the change now conflicts with the ks files for 7.

PR #1151, opened 2018-12-03
Whiteout all spaces

Sorbog:
### Description

Whiteout all spaces to reduce box size


tas50:
Good catch. That'll help depend on the partition

tas50:
@Sorbog I've reverted this since it's causing build issues. If you can figure out what platforms it works on and come up with a more targeted PR then we can move forward again.

PR #1178, opened 2019-02-17
Add template validation in Travis

tas50:
We should know we have the correct format on merge


tas50:
@cheeseplus travis seems to be busted on this repo

PR #1180, opened 2019-02-18
Updates on OpenSUSE

mcgege:
### Description

- Update all packages to newest version (similar to the other distros)
- Install some packages after adding repos as they are no longer on installer image (Leap 15)

### Issues Resolved

- Outdated software packages as no updates were installed
- In the leap-15 image the virtualbox drivers could not compile, as gcc and kernel sources have been removed from the installer dvd


tas50:
I'm glad to see  openSUSE getting some love. Thanks for putting this together

PR #1181, opened 2019-03-02
updating for 14.04.5 and 16.04.6 checksums and dl links

joshuacox:
### Description

16.04.6 and 14.04.5 have been released this fixes the old 16.04.5 and 14.04.1 references and related checksums


tas50:
I'm fairly certain the reason we didn't do this was due to failures on some virtualization platforms when we tried to build it using the latest bits. I'd have to run it through everything to check on that first. @cheeseplus might remember the details since he had to deal with all of it.

ubuntu/ubuntu-14.04-i386.json
```diff
@@ -257,12 +257,12 @@
     "headless": "",
     "http_proxy": "{{env `http_proxy`}}",
     "https_proxy": "{{env `https_proxy`}}",
-    "iso_checksum": "976044842804eafc18390505508958b559c131211160ecae5e60694bdf171f78",
+    "iso_checksum": "6071b0844ecbb8aa9a246b15c7aa20e798d42ed5b41c1b71aaea1f4283985ba4",
     "iso_checksum_type": "sha256",
-    "iso_name": "ubuntu-14.04.1-server-i386.iso",
+    "iso_name": "ubuntu-14.04.6-server-i386.iso",
```
tas50: You have 14.04.5 above and 14.04.6 here.

ubuntu/ubuntu-14.04-i386.json
```diff
@@ -257,12 +257,12 @@
...
     "memory": "1024",
     "mirror": "http://old-releases.ubuntu.com/releases",
-    "mirror_directory": "14.04.1",
+    "mirror_directory": "14.04.6",
```
cheeseplus: This looks like a typo unless there is a different patch version for i386

tas50:
14.04.5 isn't up on the archive page so I updated things to 14.04.4, but we also do the full package update anyways so the end result is a fully up to date system. I made this change in bb9404c7fed775edb70220f6b8c477458fe479e2. Thanks for pointing this out though.

PR #1182, opened 2019-03-07
Merge bento-ya back into bento

tas50:
Move all the templates into their own directory. This starts the process of shipping a binary that can be run to build and ship the boxes. This will replace Rake and will be used in Travis to validate the configs as well.


cheeseplus:
Needs rebase.

tas50:
@cheeseplus It's probably best to get this in sooner rather than later. I've used it to build the boxes. I'm sure there are issues with it and the binary needs to be updated to work with the latest / greatest bits of Vagrant cloud, but it's a step closer to having builds just work.

PR #1183, opened 2019-03-13
Amazon Linux 2

johnjelinek:
### Description

Pushes forward on the Amazon Linux 2 box.

### Issues Resolved

Continues where https://github.com/chef/bento/pull/1144 left off.


Fashaun:
Does it support vmware?

johnjelinek:
Kinda. I wrote this for the purpose of VMware, but I didn't add that part to this codebase. Part of the challenge with my VMware was seeding cloudinit with the right values to set up the interface after the template is made.

Fashaun:
Could you share your code base?

johnjelinek:
Unfortunately, I don't work for that company anymore, so I don't have it available.

PR #1187, opened 2019-04-03
remove sed -ie option command that leaves /etc/default/grube

furlongm:
For justification, see

https://github.com/sipwise/netscript/commit/4844134152e5195e3162c4df28d6fc3175866787

### Description

This change removes an option that leaves behind a file called /etc/default/grube

It was probably intended to be `sed -i -e` but only `sed -i` is required

### Issues Resolved

As above


tas50:
👍 Thanks @furlongm

PR #1193, opened 2019-04-22
Disable apt-daily earlier.

vinsonlee:
There is a race to disable apt-daily before it can trigger its own apt
process that grabs the apt lock.

Running apt-get update first adds more time for apt-daily to trigger
before it gets disabled. Then a subsequent apt command after the
disablement of apt-daily can fail grabbing the apt lock.

Move the apt-get update later and disable apt-daily earlier to give us a
better chance of winning the race.

Signed-off-by: Vinson Lee <vlee@freedesktop.org>

### Description

[Please describe what this change achieves]

### Issues Resolved

[List any existing issues this PR resolves, or any Discourse or
StackOverflow discussions that are relevant]


tas50:
Thanks @vinsonlee

PR #1196, opened 2019-05-07
Effortless Bento Boxes

jonlives:
### Description

Add Packer scripts and supporting _common scripts to support building Bento boxes pre-hardened and compliant with the DevSec baseline standards using Chef's Effortless Config and Effortless Audit habitat applications.


cheeseplus:
How can users consume these without implicitly agreeing to the new license? My fear is that there is no gating on Vagrant Cloud for a license to this forces the user into a license violating situation without their explicit knowledge.

Otherwise I'm fine with the PR but that's a pretty big one.

tas50:
Closing this out at this point since this doesn't seem like the current path forward

Issue #1199, opened 2019-06-12
Debian 9.8 does not build anymore as ISO path changed


grische:
## Version:
aa26ce8014f6b8106e82c2f7eee8e536474f751a

## Scenario:
Trying to build Debian 9.8 via packer

Some excerpt from the log:
```
==> qemu: Trying http://cdimage.debian.org/cdimage/release/9.8.0/amd64/iso-cd/debian-9.8.0-amd64-netinst.iso?checksum=sha256%3A73bd56fb4548efa384bf32fd25808b99198f81c23cf19ab9e6507d898cd58dce
==> qemu: Download failed bad response code: 404
2019/06/12 14:50:35 ui error: ==> qemu: error downloading ISO: [bad response code: 404]
==> qemu: error downloading ISO: [bad response code: 404]
2019/06/12 14:50:35 [INFO] (telemetry) ending qemu
==> qemu: leaving retrieve loop for ISO
2019/06/12 14:50:35 ui error: Build 'qemu' errored: error downloading ISO: [bad response code: 404]
```

Current path:
http://cdimage.debian.org/cdimage/release/9.8.0/amd64/iso-cd/debian-9.8.0-amd64-netinst.iso

New path should probably be:
https://cdimage.debian.org/mirror/cdimage/archive/9.8.0/amd64/iso-cd/debian-9.8.0-amd64-netinst.iso

cheeseplus:
We don't track older minor releases once we've updated so this is "resolved" by #1200  - if you still need 9.8 specifically you'll want to check our the repo at that point in time and modify or fork.

grische:
@cheeseplus #1200 is fine with me.

PR #1200, opened 2019-06-12
Update Debian 9.8 -> 9.9

tas50:
The old paths are now broken since there's a new release out


tas50:
Fixes #1199

PR #1208, opened 2019-07-08
Install rsync on popular boxes

ramereth:
Installing rsync allows users to utilize kitchen-sync [1] to speed up
test-kitchen sync tasks. Unfortunately, none of the bento boxes include rsync so
this adds it. It should had a minimal amount of disk space (less than a few MBs)

[1] https://github.com/coderanger/kitchen-sync

Signed-off-by: Lance Albertson <lance@osuosl.org>

### Description

Install rsync package on most bento boxes.

### Issues Resolved

n/a

### DIFF
--- a/packer_templates/centos/http/6/ks.cfg
+++ b/packer_templates/centos/http/6/ks.cfg
@@ -32,6 +32,7 @@ perl
 wget
 nfs-utils
 virt-what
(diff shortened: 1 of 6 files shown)


cheeseplus:
No problems here and likely overdue

PR #1221, opened 2019-07-24
 Introduce SLES 15 SP1

mattiasgiese:
- Initial commit of a config that builds SLES 15 SP1.
- Reduce footprint of resulting image.
- Add bzip2/tar packages in order to support the installation of
  VirtualBox Additions
- Misc fixes in post scripts

Signed-off-by: Mattias Giese <mattiasgiese@posteo.net>

## Description
This commit adds a recipe to build SLES 15 SP1

## Types of changes
- [x] Add new OS

## Checklist:

### DIFF
--- a/packer_templates/sles/http/sles-15-sp1-x86_64-autoinst.xml
+++ b/packer_templates/sles/http/sles-15-sp1-x86_64-autoinst.xml
@@ -1,0 +1,293 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <add-on>
+    <add_on_products config:type="list">
+      <listentry>
+        <media_url><![CDATA[dvd:///?devices=/dev/sr0]]></media_url>
+        <product>sle-module-basesystem-release</product>
+        <product_dir>/Module-Basesystem</product_dir>
+      </listentry>
+    </add_on_products>
+  </add-on>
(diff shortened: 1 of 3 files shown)


packer_templates/sles/scripts/remove-dvd-source.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh -eux
 
 zypper removerepo `zypper repos | grep 'SLES' | awk '{ print $3 }' | grep "^SLES"`;
-zypper refresh
```
tas50: I found this sped up usage of the images because the repo diff would be less between what we shipped and current.

packer_templates/sles/http/sles-15-sp1-x86_64-autoinst.xml
```diff
@@ -0,0 +1,293 @@
+<?xml version="1.0"?>
```
tas50: Is there a reason we couldn't just have 15 and 15 SP1 use the same file?

mattiasgiese: There is no reason other than to conform to the file structure of the repo as is. So surely it is possible to use this AutoYaST for both 15.0 and 15.1

packer_templates/sles/scripts/remove-dvd-source.sh
mattiasgiese: Well, keeping that line without at least adding the explicit 'exit 0' fails the build for me.

PR #1237, opened 2019-10-19
Switch to shelling out for all vagrant uploads

tas50:
The vagrant cloud gem is pretty fundamentally broken. It does not allow uploading a box that is not directly owned by the user performing the upload. Since I'm not using the bento account directly, and that seems like a horrible idea, this won't work. Instead we can just shell out to the vagrant CLI to perform the uploads. The vagrant CLI also does a much better job of this.

Other fixes:

Simplifies how we handle slugs by just using a match on the name. This means we don't have to update the slug list with the exact versions to match on every time we bump the version.
Cleans up the uploads automatically
Updates some out of date boxes
Unbreaks vmware-tools installation on RHEL
Yanks some support for FreeBSD < 11


tas50:
This is blocked by https://github.com/hashicorp/vagrant/issues/11138

tas50:
> This is blocked by [hashicorp/vagrant#11138](https://github.com/hashicorp/vagrant/issues/11138)

This issue has been solved in the vagrant API so we should be good to merge this now

PR #1255, opened 2019-11-25
Change profile to Private to avoid infinite wait

dragon788:
I've run into an infinite wait scenario when trying to build without the additional line added. While debugging I ran this script from an Administrator Powershell window and the error was always the same, the WinRM -quickconfig command doesn't seem to be getting the profile override so isn't allowing WinRM to be enabled preventing the initial Packer wait for WinRM to succeed.

## Types of changes

## Checklist:


tas50:
Nice catch. Thanks for the detailed comments. That really helps in the future.

PR #1256, opened 2019-11-27
upgrading debian 10.1 -> 10.2 and use a separate preseed cfg

pfuender:
## Description
A quick update to the Debian 10.2 release. Additionally split the preseed configuration to have one configuration per major-release (currently 8, 9, 10).

## Related Issue
-

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

### DIFF
--- a/packer_templates/debian/debian-10.2-amd64.json
+++ b/packer_templates/debian/debian-10.2-amd64.json
@@ -197,7 +197,7 @@
     }
   ],
   "variables": {
-    "box_basename": "debian-10.1",
+    "box_basename": "debian-10.2",
     "build_directory": "../../builds",
     "build_timestamp": "{{isotime \"20060102150405\"}}",
     "cpus": "1",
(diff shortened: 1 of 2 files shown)


tas50:
@pfuender was there something that requires a dedicated preseed file?

pfuender:
@tas50 - It is not strictly required (and currently it's identical to the deb9 preseed file). Personally I'm using different deb9/deb10 preseed files, that's why. 
IMHO it also makes sense, if we have different packer templates for each major release, to also keep the preseed files that way. Lmk, I can also remove the separate file.

tas50:
I'd prefer we keep the files the same unless we need to. You could rename it for 9 and 10, but the issue in the past is that someone updates something for 9 that is applicable to 10 as well and they forget to update that. Then we end up with different files and no one is sure if there's a reason for the differences or not.

pfuender:
Okay, that makes sense. I've updated it accordingly and kept the preseed files and just updated the json files.

tas50:
Thanks!

Issue #1263, opened 2019-12-19
FreeBSD-12.0-i386 on VMware Workstation: dhclient blocking 


arizvisa:
For some reason on my system, building the `FreeBSD-12.0-i386` template blocks when trying to fetch the ip address via `dhclient`. This happens when `packer` is typing in the boot sequence prior to it waiting for ssh to come up.

This will look like the following in workstation's console:
```
...
# mdmfs -s 100m md1 /tmp
random: unblocking device.
# mdmfs -s 100m md2 /mnt
# dhclient -l /tmp/dhclient.lease.em0 em0
Cannot open or create pidfile: Read-only file system
```

For some reason however this is only happening to me with `FreeBSD-12.0-i386` and \_not\_ `FreeBSD-12.0-amd64` which honestly shouldn't be the same codebase and shouldn't result in different symptoms. Nonetheless, I confirmed by trying on multiple VMware hosts all with the same result of `dhclient` blocking after displaying that error, and not backgrounding itself so the next `fetch` command can download the script for `bsdinstall`.

arizvisa:
To fix this, I simply updated the `boot_command` in each of the builders for the template and added the `-p` parameter to tell `dhclient` to write its pidfile to `/tmp`.

So where you see this line under `boot_command` (one of them uses `vtnet0`):
```
"dhclient -l /tmp/dhclient.lease.em0 em0<enter><wait>",
```

You can change it to the following. By default `dhclient` will try to write it to `/var/run/dhclient.$interface.pid`.
```
"dhclient -p /tmp/dhclient.em0.pid -l /tmp/dhclient.lease.em0 em0<enter><wait>",
```

tas50:
@arizvisa Can you submit a PR for that?

arizvisa:
Yeah, sure. I didn't write one up at first because I'm still not sure if I'm crazy because there's literally no logic change between the i386 and amd64 binaries for `dhclient`.

Should I update all of the FreeBSD templates to specify the path for the pidfile as a defensive measure?

arizvisa:
Ok. PR #1264 fixes this for me.

arizvisa:
Closed by PR #1264.

PR #1264, opened 2019-12-19
Modified the boot_command for all of the FreeBSD templates to include the pidfile path when executing dhclient.

arizvisa:
## Description
When booting up a FreeBSD template, the default "boot_command" creates two memory disks and then executes `dhclient` to setup a network interface. When setting up the interface, `dhclient` needs to write some files. One of these files is the lease file which caches the requested dhcp lease and is handled by the `-l` parameter. Another one of these files is the pidfile which stores the current pid so that it can interact with an already existing instance if necessary. Since this isn't explicitly specified, `dhclient` will use `/var/run/dhclient.$interface.pid` by default which is read-only. This seems to result in some builds blocking indefinitely while waiting for a lease.

This PR simply adds the `-p` parameter to tell `dhclient` to write its pidfile into the writable `/tmp` directory that already contains the lease cache.

## Related Issue
This closes issue #1263 for me.

## Types of changes

## Checklist:
- [X] I have read the **CONTRIBUTING** document.
- [X] All new and existing tests passed.
- [X] All commits have been signed-off for [the Developer Certificate of Origin]

### DIFF
--- a/packer_templates/freebsd/freebsd-11.3-amd64.json
+++ b/packer_templates/freebsd/freebsd-11.3-amd64.json
@@ -9,7 +9,7 @@
         "/bin/sh<enter><wait>",
         "mdmfs -s 100m md1 /tmp<enter><wait>",
         "mdmfs -s 100m md2 /mnt<enter><wait>",
-        "dhclient -l /tmp/dhclient.lease.em0 em0<enter><wait>",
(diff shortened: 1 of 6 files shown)


arizvisa:
(force pushed to include signature)

arizvisa:
Ok. good-to-go.

tas50:
Thanks for the nice description. That really helps in the future when everyone has forgotten why we do this stuff.

arizvisa:
Thx for merging this so quick.

PR #1268, opened 2020-01-06
Replace legacy service name with open-vm-tools

zestysoft:
The script was partially modified to use open-vm-tools instead of the legacy "official" vmware tools, but missed changing to the different service name that open-vm-tools uses.


tas50:
Thanks @zestysoft 

Merging this under the obvious fix policy so no DCO sign off required

PR #1286, opened 2020-05-15
oraclelinux-7: added kernelparameter for kickstart

Rendanic:
Oracle changed the behavior for the NAT network device
after rebooting the 1st time. enp3 is used during the kickstart
and eth0 after the reboot. Packer cannot connect via SSH, due
to missing network configuration for eth0.

### DIFF
--- a/packer_templates/oraclelinux/oracle-7.7-x86_64.json
+++ b/packer_templates/oraclelinux/oracle-7.7-x86_64.json
@@ -2,7 +2,7 @@
   "builders": [
     {
       "boot_command": [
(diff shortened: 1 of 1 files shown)


tas50:
@Rendanic Can you signt this for the DCO so we can accept it? Also is this something that will need to be done in 8.x as well?

Rendanic:
I will check it against OracleLinux8.

tas50:
@Rendanic if this is an issue in OL8 just open a PR for that too when you can please.

Rendanic:
This is not an issue on OL8.

PR #1299, opened 2020-07-19
Remove iso_checksum_type and references as Packer can infer them 

### LINKED ISSUE #1296: Hardcoded `iso_checksum_type` interpolation in templates prevents using file: URI for checksums
### Describe the Enhancement:
Packer is very smart and can infer the checksum type already, requiring a user/template to set it explicitly doesn't seem useful, and if they wish to they can just prefix the type into the `iso_checksum` variable itself.

https://www.packer.io/docs/builders/virtualbox/iso#iso_checksum

### Describe the Need:
When building a box via Packer that is based on a customized ISO (Windows or Linux) it can be useful to reference a local file that is generated as part of the ISO build process rather than manipulating the Packer template as part of the process. 

Allowing for the `file:` prefix of `iso_checksum` rather than interpolating the `iso_checksum_type` and `iso_checksum` in the templates makes the above possible.

### Current Alternative
I'm not sure if it is actually valid/possible to supply `iso_checksum_type: file` and `iso_checksum: some/path` because `packer validate` checks for the `iso_checksum` to have either a valid `iso_checksum_type` as a prefix (in the `variables:` block) or `file:` or nothing, so if the `iso_checksum` is just a path, without the `file:` prefix it will fail and if you do supply the full `file:some/path` the forced interpolation causes a failure to build.

### Can We Help You Implement This?:
I'll submit an MR to remove the hardcoded interpolation which shouldn't affect current functionality at all, but testing across all the OS types outside of the CI system isn't feasible for me at the moment.

dragon788:
As is they would prevent using a file: URI for the checksum which very useful if building a custom ISO with drivers/updates

Signed-off-by: dragon788 <dragon788@users.noreply.github.com>

## Description
Remove iso_checksum_type and references as Packer can infer them and they would prevent using a file: URI for the checksum which very useful if building a custom ISO with drivers/updates

## Related Issue
Fixes #1296 

## Types of changes

## Checklist:


tas50:
Thanks @dragon788

dragon788:
@tas50 I think I figured out why the issues I was seeing weren't failing the CI builds. It appears the build system has Packer 1.4.2 and not the latest 1.6.x.

PR #1361, opened 2021-05-28
Update CentOS Stream 8 to 20210524 ISO

ramereth:
Given the fact that these ISOs tend to change weekly, I sent an email to centos-devel [1] to see if they can come up with a better way to manage this in the long term for packer.

[1] https://lists.centos.org/pipermail/centos-devel/2021-May/076996.html

Signed-off-by: Lance Albertson <lance@osuosl.org>


elreydetoda:
Looks like no reply back yet for the symlink ( I just checked ), but based on [this](https://lists.centos.org/pipermail/centos-devel/2021-May/076997.html) email it you can just do an url to the checksum i.e.:

```
"iso_checksum": "file:https://mirrors.edge.kernel.org/centos/8-stream/isos/x86_64/CHECKSUM"
```

I just tried it out on one of the projects I am working on, and it worked. So, that will at least solve one of the issues. Then you will just have to place the URL for the updated iso ( until/if they add a symlink).

ramereth:
@elreydetoda they replied today and seemed as though they're willing to add a "latest" symlink. Once they do that, I'll update this to use the method above for the checksums.

ProfessorManhattan:
If they don't add it you can check out this Python module i had someone
make:

https://pypi.org/project/latestos/

ramereth:
For those following, I created an [upstream bug](https://bugzilla.redhat.com/show_bug.cgi?id=1966812) to track this further.

PR #1441, opened 2022-12-11
Make sure the uploaded dir is in git

tas50:
The bento bin fails without this directory


Stromweld (approved):
LGTM

PR #1442, opened 2022-12-11
Remove EOL FreeBSD 12.2/13.0 and add 12.4

tas50:
Update to the latest things that are supported


packer_templates/freebsd/freebsd-13.0-amd64.json
```diff
@@ -1,211 +0,0 @@
-{
```
Stromweld, line 1: Shouldn't this be updated to 13.1 instead of deleted

Stromweld: nvm just looked and 13.1 file is already there.

Stromweld (approved):
LGTM

Stromweld (approved):
LGTM

PR #1449, opened 2022-12-12
Fix failing Amazon Linux builds

tas50:
The http variable without the dir causes failures and the instructions were a tiny bit wrong


Stromweld (approved):
LGTM

PR #1450, opened 2022-12-12
Remove extra checks from alma/rocky scripts

tas50:
Now that these are unique to Alma/Rocky remove all the junk for old centos, redhat, and oracle


Stromweld (approved):
LGTM

PR #1487, opened 2023-03-08
fix(ubuntu-20.04): boot from media works again

jhalfmoon:
## Description

Fixed Ubuntu
## Types of changes

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.
- [ ] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).


jhalfmoon:
GH Actions is breaking with 'Secret PACKER_GITHUB_API_TOKEN is required, but not provided while calling."

jhalfmoon:
Note that having a boot_wait  2s or less is critical to getting into the Ubuntu 20.04 bootmenu, as it has a time-out of 2 seconds. A boot_wait of more that 2s will make prevent the packer from accessing the boot menu.

Stromweld:
gotcha thanks for the heads up. Ill add that back. trying to figure out this github_token thing for the packer init. Apparently there is a limit unless you provide a token but the configured secrets don't work for forks.

jhalfmoon:
Note: The escape characters in the boot_command break qemu build.

Stromweld:
Sorry about that, some builds fail without that and others work, still baffling me why some need the escape characters. Looking back at the old json templates you are correct that 20.04 didn't have them but 22.10 do.

Stromweld:
I reverted the ci-build.yml, currently running all builds to verify which builds work on which virtual machine provider.  Currently they all are running on 3 mac-mini hosts 1 intel mac mini with 4 runners configured on it, 1 M2 mac mini with 5 runners configured on it, and 1 older mac mini that is running windows 11 OS and hyper-v with 2 runners configured on it. This is temporary until we can build out our cloud machines as runners. Once all the builds are working the idea would be to only run a build for each provider in parallel only for each pkrvars.hcl file added or modified. Then once a quarter all builds would run to refresh boxes in vagrant quarterly with latest patches.

jhalfmoon:
On a side-note: My apologies for the force-pushes ; I'm very much used to working solo on feature-branches and force-push is not an issue in such situations. In this case I'll try to remember that I'm not alone on this branch.

jhalfmoon:
@Stromweld Regarding the CI builds; The current build cycle takes over 18 hours. I would recommend making the build-all-in-parallel workflow a manually triggered workflow. For pull requests like this you would be better of saving many hours of CPU cycles, waiting and electrical energy by using a more serialized approach, for example as per the workflow I previously committed. It will make for a much shorter fix / test cycle.

Stromweld (approved):
Looks good. Thanks for the PR and letting me test fixes for outside PR's and the secrets/github_token issue for the api calls.

Stromweld:
No worries on the force push. Normally I wouldn't piggy back on someone else's PR but since I couldn't test the secrets issue for non collaborators without creating a dummy account this was the first time I had seen that issue and wanted to fix it right away.

I did add the concurrency now that I have that cleanup job. Since these are self hosted nodes when jobs got cancelled I had to manually remove the VM's from the provider that was in the middle of a build otherwise next run would run into an error on vm with same name already existing.

PR #1490, opened 2023-03-13
Match any kernel version in debian purge command

upekkha:
The `cleanup_debian.sh` script was only removing kernel versions starting with `[234]` and thus not matching the current kernel versions 5 or 6. This PR suggests to match any version `[1-9]`.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [X] I have read the **CONTRIBUTING** document.
- [X] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### DIFF
--- a/packer_templates/scripts/debian/cleanup_debian.sh
+++ b/packer_templates/scripts/debian/cleanup_debian.sh
@@ -9,7 +9,7 @@ dpkg --list \
 echo "remove specific Linux kernels, such as linux-image-4.9.0-13-amd64 but keeps the current kernel and does not touch the virtual packages"
 dpkg --list \
     | awk '{ print $2 }' \
-    | grep 'linux-image-[234].*' \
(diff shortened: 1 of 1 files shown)


Stromweld:
Have you tested this change? My only concern is that it'll remove all linux-image packages even for the current kernel.

upekkha:
I locally tested with Debian 10, 11 and upcoming 12. The currently running kernel is anyway excluded by the `grep -v "$(uname -r)"` on the next line.

Stromweld:
gotcha I see it now. makes sense thanks.

Stromweld (approved):
LGTM

PR #1526, opened 2023-09-06
Keep linux-kernel-header package installed for Parallels builds

mpdude:
When using [this Ubuntu Vagrant box](https://app.vagrantup.com/bento/boxes/ubuntu-22.04/versions/202303.13.0) with Parallels Desktop, the `vagrant-parallels` plugin upon boot detects  that the version of Parallels Tools installed in the box is outdated and tries to update it.

The problem is that a due re-compilation of Parallels Tools depends on `linux-headers-5.15.0-67-generic`, with the package name/version matching the version of the currently running kernel.

However, the Kernel version included in the box image has since then replaced with a newer version (in the upstream Ubuntu repositories), and it seems the matching header files can no longer be installed. They probably came from an Ubuntu security update release channel where the package has been replaced with a newer version since the box was built.

Full details can be found in the issue at https://github.com/Parallels/vagrant-parallels/issues/458.

By keeping the `linux-headers-*` package during/after installation, also a freshly booted box should be able to perform the update. The package is, in fact, installed automatically when Parallels Tools are added, but after that is is removed by the `cleanup_ubuntu.sh` script.

### DIFF
--- a/packer_templates/pkr-builder.pkr.hcl
+++ b/packer_templates/pkr-builder.pkr.hcl
@@ -115,6 +115,7 @@ locals {
             "${path.root}/scripts/_common/parallels.sh",
             "${path.root}/scripts/${var.os_name}/hyperv_${var.os_name}.sh",
             "${path.root}/scripts/${var.os_name}/cleanup_${var.os_name}.sh",
+            "${path.root}/scripts/_common/parallels_post_cleanup_debian_ubuntu.sh",
             "${path.root}/scripts/_common/minimize.sh"
             ] : (
             var.os_name == "fedora" ? [
--- a/packer_templates/scripts/_common/parallels_post_cleanup_debian_ubuntu.sh
+++ b/packer_templates/scripts/_common/parallels_post_cleanup_debian_ubuntu.sh
@@ -1,0 +1,14 @@
+#!/bin/sh -eux
+
+case "$PACKER_BUILDER_TYPE" in
+parallels-iso|parallels-pvm)
+    # This runs in a special script after cleanup_{ubuntu|debian}.sh, which cleans up the kernel headers
+    # in general (we want to keep them for Parallels Desktop only).
+
+    echo "Installing linux-kernel-headers for the current kernel version, to allow re-compilation of Parallels Tools upon boot"
+
+    if [ -f "/usr/bin/apt-get" ]; then
+        apt-get install -y linux-headers-"$(uname -r)"
+    fi
+    ;;
+esac


Stromweld (approved):
LGTM

mpdude:
Will a merge also lead to a new box being built and published?

mpdude:
Did this PR break the GHA workflows?

Stromweld:
No they are running. With the switch to hcl2 there is still more tweaking and testing of various builds that need to happen. Most probably need a variable for the boot_wait per provider and tweaks to the os_pkrvariables for each OS to get them to work.

mpdude:
You need #1527 first to fix outdated Ubuntu 22.04 ISO URLs

mpdude:
😱 Did this really necessitate to run more than 100 VM image builds, spending more than 17h in CI?

The ecological footprint 🌱 is devastating 🤔

Stromweld:
eventually I want to add logic to only build the pkvars files that have changed. in the changelog there is a list of things to do with this repo. I'm the only one working on it on the side. Right now all the builds happen on 3 mac mini's running in my basement hosting the build runners.

mpdude:
😩

sorry if that sounded like criticism, that was not my intention.

I know almost nothing about packer and the stuff you’re doing here… not sure, you think I can help?

Would it be possible to re-organize files in a way that make it clear which VM images depend on changes in which directories?

mpdude:
You can limit GHA workflows to run only when files in a directory changed. So, when there were directories for files that affect all VMs of a certain distro, or all of a certain hypervisor etc, maybe that mechanism could be used?

Stromweld:
It's all good. That approach would still build a lot of boxes. I've done something similar here to only run json-lint job if a json file has been added or modified. Simply need to get the list of files as output from the previous job and use it as input to the build matrix. https://github.com/Stromweld/github-workflows/blob/main/.github/workflows/json-lint.yml may need to run custom script to parse them out based on cpu architecture.

mpdude:
I can confirm the initial issue seems to be fixed in https://app.vagrantup.com/bento/boxes/ubuntu-22.04/versions/202309.08.0 – at least, no Parallels Tools update problem at boot (no surprise, it's all fresh and shiny and new) and the box comes with the header files installed.

Stromweld:
Thanks, real test will be the next update for parallels. These new machines were built with the latest parallels 19.0.0 version so there shouldn't be a tools upgrade.

mpdude:
Should the v19 tools also work with previous versions of Parallels?

Stromweld:
from my understanding yes. I haven't tested and verified though.

shir:
@mpdude  will the ubuntu-22.04-arm64 box updated too to fix the issue?

mpdude:
@shir Generally yes, but I don't know if that image has been successfully rebuilt and published at all

shir:
@mpdude the latest released version for arm64 is v202306.30.0 and it doesn't have this fix. https://app.vagrantup.com/bento/boxes/ubuntu-22.04-arm64

gndk:
@Stromweld Could you please build/publish a new version of the [ubuntu-22.04-arm64](https://app.vagrantup.com/bento/boxes/ubuntu-22.04-arm64) box for parallels provider with this fix? Thank you!

Stromweld:
yes I'll work on that now.

Stromweld:
this should be up there now

gndk:
Great, thanks again!

PR #1527, opened 2023-09-07
Update Ubuntu 22.04 `iso_url`

mpdude:
The https://releases.ubuntu.com/jammy/ubuntu-22.04.2-live-server-amd64.iso address is no longer available (404). Since August 2023, there is a .3 release.


Stromweld (approved):
LGTM

PR #1614, opened 2025-05-13
Fix: prevent cleanup script crash when no packages are found

akelbikhanov:
This PR fixes a bug in the cleanup_debian.sh and cleanup_ubuntu.sh scripts where certain dpkg | grep | xargs apt-get purge pipelines may fail if no matching packages are found.

When running on minimal images (e.g. with headers or dev packages already removed), grep returns non-zero code and causes the script to exit under set -e. Additionally, xargs without -r may invoke apt-get with no arguments, which leads to an error.

Changes:
- added '|| true' after all grep calls to avoid pipeline failures
- added xargs '-r' to skip apt-get if the input is empty

These changes are safe, maintain existing logic, and ensure the scripts work reliably in clean or partially pre-cleaned environments.

### DIFF
--- a/packer_templates/scripts/debian/cleanup_debian.sh
+++ b/packer_templates/scripts/debian/cleanup_debian.sh
@@ -3,27 +3,27 @@
 echo "remove linux-headers"
 dpkg --list \
   | awk '{ print $2 }' \
-  | grep 'linux-headers' \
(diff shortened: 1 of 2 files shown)


Stromweld (approved):
LGTM

PR #1615, opened 2025-05-13
fix: wrap late_command in sh -c to ensure proper redirection

akelbikhanov:
This PR fixes an issue in the preseed.cfg where multiple d-i preseed/late_command entries silently override each other, causing only the last one to execute.

Additionally, shell constructs like > (redirection) and && are not reliably interpreted by the installer unless explicitly passed to sh -c.

Changes:
- Combined multiple late_command lines into a single command chain
- Wrapped the full command in sh -c "..." to ensure correct shell parsing and file redirection

This ensures all late-stage installation tasks (like configuring sudo access and modifying APT sources) are executed consistently and predictably.


Stromweld (approved):
LGTM

Issue #1628, opened 2025-09-03
Debian libvirt images


votdev:
# Describe the Enhancement

Add support for Debian, e.g. Trixie/v13, libvirt images.

## Describe the Need

Since the Debian project itself [no longer provides images](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1110834), the possibility of obtaining Debian images from reputable sources is now unfortunately very limited.

## Current Alternative

Use existing images from unknown sources.

## Can We Help You Implement This

No 🙁

csenger:
The broken Debian builds are due to libvirt using virtio which names block devices `/dev/vda`. The configuration is hard coded to use `/dev/sda`. I created a PR with the changes (#1631). 

@votdev Until there is a bento release with a libvirt box, you can build your own from the bento repository. Install bento as described in the README. Clone the repository and apply the changes that replace `/dev/sda`, like in #1631, and run `bento build...` from the repository root. `bento build -d --gui -o qemu.vm --vars "ssh_timeout=90m" os_pkrvars/debian/debian-13-x86_64.pkrvars.hcl ` worked for me and opens VNC to follow the progress/see potential errors). Then you can `vagrant box add <your box name> <box file>` locally. Or throw it on a server or upload it to hashicorp cloud.

PR #1631, opened 2025-09-24
Fix Debian libvirt build

csenger:
This fixes the Debian libvirt builds by using `grub-installer/bootdev` default so it works with libvirt's `/dev/vda` and `/dev/sda` in other builds.

## Description

The libvirt build fails with

```
Unable to install grub in /dev/sda
Executing 'grub-install /dev/sda' failes.

This is a fatal error
```
The libvirt virtio driver creates the disk as `/dev/vda`.

This sets `grub-installer/bootdev` to default, which internally is "(hd0)", instead of hard coding a device which picks the first disk device. It works with either vda or sda. Addresses #1628

Debian's "Debian GNU/Linux Installation Guide", Section [B.4.12. Boot loader installation](https://www.debian.org/releases/stable/amd64/apbs04.en.html#preseed-bootloader), states: "Due notably to potential USB sticks, the location of the primary drive can not be determined safely in general, so this needs to be specified". This should be no problem as there is no second disk device.

An alternative approach would be to use a preseed/early_command to create a symlink if `/dev/sda` is missing. `grub-install` works fine with a symlink. I'm not sure if this would require to fix the grub configuration at the end of the preseed for later grub updates.

Debian 12 and 13 builds work with libvirt and virtualbox on x86, tested manually. I could not get `qemu-system-aarch64` to boot the iso (host is Ubuntu 24.04/x86) to test the aarch64 build. I do not have access to other providers.

`bento test`: I did not find good documentation what and how to set up to test everything with testkitchen.  

* bento test passes for virtualbox. 
* libvirt fails with
  ```
  ==> default: Preparing to edit /etc/exports. Administrator privileges will be required...
  STDERR: [fog][WARNING] Unrecognized arguments: libvirt_ip_command
  Vagrant failed to install an updated NFS exports file. This may be
  due to overly restrictive permissions on your NFS exports file. Please
  validate them and try again.

  command: sudo chown 0:0 /tmp/vagrant-exports
  stdout: 
  stderr: sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
  sudo: a password is required
  ---- End output of vagrant up --no-provision --provider libvirt ----
  Ran vagrant up --no-provision --provider libvirt returned 1] on default-debian-1212-amd64-libvirt
  >>>>>> ----------------------
  ``` 
  The libvirt_ip_command is only a warning (from a vagrant-libvirt dependency IIRC, no update for vagrant-libvirt since). 
  The NFS error happened in my normal Vagrant-Setup too and where I have NFS disabled now. The test leaves a running VM behind.
* qemu (after installing the vagrant-qemu plugin) fails. No idea how to fix this in testkitchen.
  ```
       Bringing machine 'default' up with 'qemu' provider...
       Invalid config.
       
       Error: Invalid qemu dir: /opt/homebrew/share/qemu
  ```

## Related Issue

None

## Types of changes

## Checklist:

- [ ] If `Gemfile.lock` has changed, I have used `--conservative` to do it and included the full output in the Description above.


Stromweld (approved):
nice find, thanks for the fix

PR #1673, opened 2026-03-17
Fix vmware-iso validation by explicitly setting tools_mode

t-mtsmt:
## Description

This change fixes `vmware-iso` validation when VMware tools upload settings are configured.

Bento already sets `tools_upload_flavor` and `tools_upload_path` for VMware builds, but does not set `tools_mode`. With current VMware builder validation, `tools_mode` must be explicitly specified whenever tools configuration is present.

Without this change, validation fails with the following error:

```text
Error: 1 error(s) occurred:

* 'tools_mode' must be explicitly specified when using any tools configuration

  on packer_templates\pkr-sources.pkr.hcl line 547:
  (source code not available)
```

This PR adds a `vmware_tools_mode` variable, derives a default value from the existing upload configuration, and passes `tools_mode` to the `vmware-iso` source.

The default behavior remains the same:
- use `upload` when a VMware tools upload flavor is configured
- use `disable` when no upload flavor is available
- allow explicit override through `vmware_tools_mode`

## Related Issue

No related issue.

## Types of changes

## Checklist:

- [ ] If `Gemfile.lock` has changed, I have used `--conservative` to do it and included the full output in the Description above.

### DIFF
--- a/packer_templates/pkr-sources.pkr.hcl
+++ b/packer_templates/pkr-sources.pkr.hcl
@@ -211,6 +211,9 @@ locals {
       var.os_name == "macos" ? "darwin" : null
     )
   ) : var.vmware_tools_upload_flavor
+  vmware_tools_mode = var.vmware_tools_mode == null ? (
+    local.vmware_tools_upload_flavor == null ? "disable" : "upload"
+  ) : var.vmware_tools_mode
   vmware_tools_upload_path = var.vmware_tools_upload_path == null ? (
     var.is_windows ? "c:\\vmware-tools.iso" : "/tmp/vmware-tools.iso"
   ) : var.vmware_tools_upload_path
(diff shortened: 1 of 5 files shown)


Stromweld:
now that they have attach fixed I think this should default to attach. upload was only used as it was the only one supported for all platforms. Attach speeds things up since it doesn't have to spend the time copying the iso to the server.

packer_templates/pkr-sources.pkr.hcl
```diff
@@ -211,6 +211,9 @@ locals {
       var.os_name == "macos" ? "darwin" : null
     )
   ) : var.vmware_tools_upload_flavor
+  vmware_tools_mode = var.vmware_tools_mode == null ? (
+    local.vmware_tools_upload_flavor == null ? "disable" : "upload"
+  ) : var.vmware_tools_mode
```
Stromweld: I'd prefer to default to attach. Upload drags out the build and takes a long time. Previously they didn't have the option to attach.

